PR #80, opened 2020-07-27
feat: support OR filter

### LINKED ISSUE #44: Implement and() & or()
[Reference](http://postgrest.org/en/v7.0.0/api.html#horizontal-filtering-rows).
- `and()` & `or()` directly refer to the conditional statements AND & OR respectively.
- Although functionality for `and()` already exists through chaining functions, the use case for it would be to allow complex chaining of both `and()` and `or()` together.

soedirgo:
Adds support for OR filter like so:

```js
client.from('countries').or('id.gt.20,and(name.eq.New Zealand,name.eq.France)')
```

Support for AND is not added because they're redundant with chaining filters.

Closes #44.

### DIFF
--- a/src/Builder.js
+++ b/src/Builder.js
@@ -45,6 +45,10 @@ class Builder {
           request.not(queryFilter.columnName, queryFilter.operator, queryFilter.criteria)
           break
 
+        case 'or':
+          request.or(queryFilter.filters)
+          break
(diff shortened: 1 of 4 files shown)


kiwicopple:
Amazing work @soedirgo 🌮

PR #109, opened 2020-09-28
fix: change fetch polyfill

soedirgo:

Bug fix.

There's [some problems](https://codesandbox.io/s/compassionate-mayer-5nk8h) using postgrest-js from jspm, probably related to https://github.com/developit/unfetch/issues/104 and https://github.com/developit/unfetch/pull/109. The repo doesn't look too healthy, so I'm switching to [cross-fetch](https://github.com/lquixada/cross-fetch) for the fetch polyfill, which uses [github/fetch](https://github.com/github/fetch) under the hood.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1395,6 +1395,14 @@
       "integrity": "sha1-tf1UIgqivFq1eqtxQMlAdUUDwac=",
       "dev": true
     },
+    "cross-fetch": {
+      "version": "3.0.6",
+      "resolved": "https://registry.npmjs.org/cross-fetch/-/cross-fetch-3.0.6.tgz",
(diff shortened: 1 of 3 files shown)


soedirgo:
Update: Apparently the bug [doesn't exist](https://codesandbox.io/s/cold-cache-3ilge) when using [Skypack](https://www.skypack.dev), but I'll keep this change for now.

kiwicopple:
Looks like they finally solved this @soedirgo https://github.com/developit/unfetch/pull/109

I want to confirm that we'll stick with cross-fetch? If so, I will migrate gotrue-js too

soedirgo:
Yeah, I'm partial towards cross-fetch because of project health.

Issue #78, opened 2020-07-20
Implement `on_conflict` 


awalias:
Implement `on_conflict` option on inserts

from: [http://postgrest.org/en/latest/api.html#on-conflict](http://postgrest.org/en/latest/api.html#on-conflict)

The user should be able to upsert on UNIQUE columns (that may not be the primary key)

kiwicopple:
@soedirgo is this one still relevant after all your changes? If so, perhaps we label it for Hacktoberfest and see if the community want to contribute?

soedirgo:
👍 Sounds good, `insert` is currently missing this feature. This can be an [option](https://github.com/supabase/postgrest-js/blob/d2f944a347a86f79397ca26876effb39d356241c/src/builder.ts#L122) alongside `upsert` that gets used when `upsert` is `true`.

PR #113, opened 2020-10-02
Feat: implement on_conflict

izqalan:

Added `on_conflict` feature for `insert` method.

Missing `on_conflict` feature for `insert` method. https://github.com/supabase/postgrest-js/issues/78

## What is the new behavior?

When inserting, if `upsert: true` and _UNIQUE_ column specified on `on_conflict: column` option set `on_conflict` searchParams.

## Additional context

http://postgrest.org/en/v7.0.0/api.html?highlight=on%20conflict#on-conflict


src/builder.ts
```diff
@@ -117,12 +117,17 @@ export class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
    *
    * @param values  The values to insert.
    * @param upsert  If `true`, performs an UPSERT.
+   * @param on_conflict By specifying the `on_conflict` query parameter, you can make UPSERT work on a column(s) that has a UNIQUE constraint.
    */
-  insert(values: Partial<T> | Partial<T>[], { upsert = false } = {}): PostgrestBuilder<T> {
+  insert(
+    values: Partial<T> | Partial<T>[],
+    { upsert = false, on_conflict }: { upsert?: boolean; on_conflict?: string } = {}
```
soedirgo, line 124:
```suggestion
    { upsert = false, onConflict }: { upsert?: boolean; onConflict?: string } = {}
```

test/basic.test.ts
```diff
@@ -29,6 +29,16 @@ test('switch schema', async () => {
...
+test('on_conflict insert', async () => {
+  let res = await postgrest
+    .from('users')
+    .insert([
+      { username: 'izqalan', catchphrase: 'bat', status: 'OFFLINE', age_range: [20, 25]},
+      { username: 'dragarcia', catchphrase: "'fat' 'cat'", status: 'OFFLINE', age_range: [20, 30]}
+    ], { upsert: true , on_conflict: 'username'});
+  expect(res).toMatchSnapshot()
+})
+
```
soedirgo, line 41: I think you can make a minimal test with something like:

```js
const res = await postgrest
  .from('users')
  .insert({ username: 'dragarcia' }, { upsert: true, onConflict: 'username' })
expect(res).toMatchSnapshot()
```

This preserves the state of the database, leaving other tests unaffected. In an ideal world, each test would be independent from each other, but this is the best we can come up with right now.

src/builder.ts
```diff
@@ -117,12 +117,17 @@ export class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
...
+  ): PostgrestBuilder<T> {
     this.method = 'POST'
     this.headers['Prefer'] = upsert
       ? 'return=representation,resolution=merge-duplicates'
       : 'return=representation'
+    if (upsert === true && on_conflict !== undefined) this.url.searchParams.set('on_conflict', on_conflict)
```
soedirgo, line 130:
```suggestion
    if (upsert && on_conflict !== undefined) this.url.searchParams.set('on_conflict', on_conflict)
```

Minor nit :)

src/builder.ts
soedirgo, line 124: `onConflict` would be a better name, following JS's naming convention.

soedirgo (changes requested):
Hey again @izqalan! 👋

I just have a couple of nits while this is in a draft. Otherwise, looks good!

test/basic.test.ts
izqalan: ah, that's better

izqalan:
@soedirgo updated with your suggestions.

soedirgo:
Awesome, thanks again @izqalan!

TomasHubelbauer:
Seems like this `insert` overload has since been deprecated (judging by the TypeScript types), use `upsert` like this:
```typescript
await supabase.from('ips').upsert({ updated_at: new Date(), ip: request.ip, geo }, { onConflict: 'ip' });
```

magick93:
How would I translate the following sql insert to the supabase js client?

```
INSERT INTO comment_votes (comment_id, user_id, vote)
VALUES('e9a8b088-aa2f-45ac-b961-e5b71cb5ef42','f1b80dc9-5830-432a-b137-d3c6b9ba2b4b', null) 
ON CONFLICT ON CONSTRAINT comment_votes_pkey
DO 
   UPDATE SET vote = excluded.vote;
```

Can I specify a composite key as a constraint for the conflict?

steve-chavez:
@magick93 Replied on https://github.com/supabase/supabase/discussions/8093

PR #117, opened 2020-10-12
Allow filters and transforms on insert

soedirgo:

Feature.

## What is the new behavior?

Allow vertical filters and transforms on the result of `insert()`. E.g.:

```js
const { error, data } = await postgrest
    .from('users')
    .insert({ ... })
    .single()
```

Most of the time, we just want a single object for the above.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -2,7 +2,7 @@ import fetch from 'cross-fetch'
 
 /**
  * Error format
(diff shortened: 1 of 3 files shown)


kiwicopple:
Perfect, thanks Bobbie. I'll merge this in myself so I can keep going on supabase-js

PR #119, opened 2020-10-19
refactor: proposal to refactor the library with some additional exports

kiwicopple:
## Refactor

- added some additional exports which I need in `supabase-js`
- Refactored the code - this is an opinion, interested to get your thoughts @soedirgo. Basically I move each of the key classes into their own file - more consistent with the other libraries. Is there any reason it was originally all in one `builder.js` file?


soedirgo (approved):
LGTM!

> Basically I move each of the key classes into their own file

Yep, this is fine. I thought about splitting off filters but was worried about cyclic imports--seems like it's not an issue.

PR #132, opened 2020-11-12
Fix ordering on JSON columns

### LINKED ISSUE #131: Ordering JSON column not working
# Bug report

## Describe the bug

This is failing 

```
const { data: products, error } = await supabase
        .from('products')
        .select('*')
        .order('metadata->index'
```

Caused by this: 

https://github.com/supabase/postgrest-js/blob/43b093d7eaab1c36b76b11cbcb1cfc2b3605c4a2/src/lib/PostgrestTransformBuilder.ts#L27

If we remove the quotes from the `"${column}"` then it works.

soedirgo:

Bug fix.

## What is the new behavior?

Previously, `order` columns wrap the column name in quotes to allow
characters like `.` without having the user wrap it manually, but this
breaks on JSON columns, so we're removing the quotes.

## Additional context

Closes #131.


kiwicopple:
Niice @soedirgo

PR #146, opened 2020-12-31
feat: add vertical filtering on CUD

### LINKED ISSUE #143: select columns from an insert
# Feature request

## Is your feature request related to a problem? Please describe.
I need a way to perform an upsert, and know if I performed an update or an insert afterwards. Postgrest has an open issue to support this via 200 and 201 status codes https://github.com/PostgREST/postgrest/issues/1070. For the time being though, I can infer this on the client side using the `xmax` column. E.g.
```
POST http://<supabase-db-url>.supabase.co/rest/v1/items?select=id,xmax
```
If `xmax === "0"` then the row was inserted, otherwise an update occurred. Unfortunately, there is no way to write this kind of query with the supabase client.

## Describe the solution you'd like
```js
const result = await supabase
  .from('items')
  .insert([row], { onConflict: 'unique_field', upsert: true })
  .select('id,xmax')
if (result.data[0].xmax === "0") {
  console.log('insert')
} else {
  console.log('update')
}
```

## Describe alternatives you've considered
write an rpc function to wrap this logic

## Additional context
an issue on the PostgREST repo discussing possible solutions here: https://github.com/PostgREST/postgrest/issues/1683

selects on insert is valuable for multiple reasons outside of my outlined use case. Any performant client will be interested in returning the minimal rows necessary after an update/insert, rather than all of them all the time. I feel like I saw an issue or discussion around the default behavior around the `RETURNING` clause but for the life of me I cannot find it.

soedirgo:

Feature.

Can't do vertical filtering on INSERT/UPDATE/DELETE.

## What is the new behavior?

Allow vertical filtering (`select`) during transform (i.e. at the end of the method chain):
```js
await postgrest
  .from('items')
  .update({ quantity: 1 })
  .eq('category', 'fruits')
  .select('id,xmax')

await postgrest
    .rpc('get_username_and_status', { name_param: 'supabot' })
    .select('status')
```

## Additional context

Closes #143.

### DIFF
--- a/src/PostgrestClient.ts
+++ b/src/PostgrestClient.ts
@@ -1,4 +1,5 @@
 import PostgrestQueryBuilder from './lib/PostgrestQueryBuilder'
+import PostgrestTransformBuilder from './lib/PostgrestTransformBuilder'
 import { PostgrestBuilder } from './lib/types'
 
 export default class PostgrestClient {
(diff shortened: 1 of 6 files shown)


kiwicopple:
This is awesome @soedirgo . No breaking changes here right? Merge when you're ready 👍

Issue #94, opened 2020-09-07
feat: Support COUNT of relevant rows


soedirgo:
# Feature request

## Is your feature request related to a problem? Please describe.

I'd like to perform `COUNT(*)` on relevant rows.

## Describe the solution you'd like

A `PostgrestTransformBuilder` method. PostgREST [supports this](https://postgrest.org/en/stable/api.html#exact-count).

stupidawesome:
Postgrest reports counts in the `Content-Range` header. This has to be requested by setting `Prefer: count={strategy}` where `{strategy}` is one of `exact`, `planned` or `estimated`.

`range()` could be modified to accept a `count` option as this is useful for displaying search results with a paginator. 

```ts
range(
        from: number,
        to: number,
        {
            count = "estimated",
            foreignTable,
        }: {
            foreignTable?: string
            count: "exact" | "estimated" | "planned"
        } = { count: "estimated" },
    ): PostgrestTransformBuilder<T> {
        const keyOffset =
            typeof foreignTable === "undefined"
                ? "offset"
                : `"${foreignTable}".offset`
        const keyLimit =
            typeof foreignTable === "undefined"
                ? "limit"
                : `"${foreignTable}".limit`
        this.url.searchParams.set(keyOffset, `${from}`)
        // Range is inclusive, so add 1
        this.url.searchParams.set(keyLimit, `${to - from + 1}`)
        this.headers["Prefer"] = `count=${count}`
        return this
    }
```

Then in the fetch promise could return this as part of the response:

```ts
const contentRange = res.headers.get("Content-Range")
if (contentRange) {
    const parts = contentRange.split(/[-\/]/)
    // Note: this doesn't not handle responses like `0/*`
    range = {
        from: parseInt(parts[0], 10) + 1, // zero-based index
        to: parseInt(parts[1], 10) + 1, // zero-based index
        total: parseInt(parts[2], 10),
    }
}

return {
    range,
    data,
    status,
    ...etc
}
```

This would involve modifying the [response format](https://github.com/supabase/supabase-js/issues/32#issuecomment-704059049)

soedirgo:
Hey @stupidawesome! This issue is for getting just the `COUNT` using a `HEAD` request, as laid out [here](http://postgrest.org/en/v7.0.0/api.html#exact-count).

I'd like to keep the library simple, so we don't plan to expose the HTTP response headers.

dshukertjr:
[...]

PR #147, opened 2021-01-09
Adding head and count option on select(), insert(), update(), delete(), and rpc()

dshukertjr:

It will add `head` and `count` option to allow querying row count. 

Cannot get row count. 

## What is the new behavior?

Enable retrieving row count. 

## Related issues

#94

### DIFF
--- a/src/lib/PostgrestQueryBuilder.ts
+++ b/src/lib/PostgrestQueryBuilder.ts
@@ -69,14 +69,11 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
       upsert = false,
       onConflict,
       returning = 'representation',
+      count = null,
     }: {
       upsert?: boolean
       onConflict?: string
       returning?: 'minimal' | 'representation'
-    } = {},
-    {
-      count = null,
-    }: {
       count?: null | 'exact' | 'planned' | 'estimated'
     } = {}
   ): PostgrestFilterBuilder<T> {
@@ -90,7 +87,8 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
     if (upsert && onConflict !== undefined) this.url.searchParams.set('on_conflict', onConflict)
     this.body = values
     if (count) {
-      this.headers['Prefer'] = `count=${count}`
+      prefersHeaders.push(`count=${count}`)
+      this.headers['Prefer'] = prefersHeaders.join(',')
     }
     return new PostgrestFilterBuilder(this)
   }
@@ -103,18 +101,22 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
    */
   update(
     values: Partial<T>,
-    { returning = 'representation' }: { returning?: 'minimal' | 'representation' } = {},
     {
+      returning = 'representation',
       count = null,
     }: {
+      returning?: 'minimal' | 'representation'
       count?: null | 'exact' | 'planned' | 'estimated'
     } = {}
   ): PostgrestFilterBuilder<T> {
     this.method = 'PATCH'
-    this.headers['Prefer'] = `return=${returning}`
+    let prefersHeaders = []
+    prefersHeaders.push(`return=${returning}`)
+    this.headers['Prefer'] = prefersHeaders.join(',')
     this.body = values
     if (count) {
-      this.headers['Prefer'] = `count=${count}`
+      prefersHeaders.push(`count=${count}`)
+      this.headers['Prefer'] = prefersHeaders.join(',')
     }
     return new PostgrestFilterBuilder(this)
   }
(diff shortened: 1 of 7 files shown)


dshukertjr:
I have implemented `count` option for select, and wanted some feedback so that what I am doing here is along the line of what discussed [here](https://github.com/supabase/postgrest-js/issues/94).

dshukertjr:
@steve-chavez  @kiwicopple  @soedirgo 
I am done implementing the `count` and `head` option on `select`, and added `count` field to the test snapshot at this point. If everything looks fine, I would move on to implementing `count` and `head` to `insert()`, `update()`, `delete()`, and `rpc()`.

src/lib/types.ts
```diff
@@ -80,18 +82,36 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
...
-          const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
-          data = isReturnMinimal ? null : await res.json()
+          if (this.method !== 'HEAD') {
+            const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
+            data = isReturnMinimal ? null : await res.json()
+          } else {
+            data = null
+          }
+
+          const countHeader = this.headers['Prefer']?.match(/count=(exact|planned|estimated)/)
```
dshukertjr, line 95: I feel like there is a better way of extracting the `count` value from the header here. Any suggestions would be great!

src/lib/PostgrestQueryBuilder.ts
```diff
@@ -17,8 +17,13 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
    * Performs vertical filtering with SELECT.
    *
    * @param columns  The columns to retrieve, separated by commas.
+   * @param head  When set to true, select will void data.
+   * @param count  Count algorithm to use to count rows in a table.
    */
-  select(columns = '*'): PostgrestFilterBuilder<T> {
+  select(
+    columns = '*',
+    { head, count }: { head?: boolean; count?: 'exact' | 'planned' | 'estimated' } = {}
```
soedirgo, line 25: Might be better to explicitly set the default values here:
```suggestion
    {
      head = false,
      count = null,
    }: {
      head?: boolean
      count?: null | 'exact' | 'planned' | 'estimated'
    } = {}
```

src/lib/types.ts
soedirgo: I think this is fine for now. An alternative would be to add a `prefers` object property to `PostgrestBuilder` and join everything in `then()`.

soedirgo (commented):
Nice work @dshukertjr, just one nit :)

Also, if you could add a simple test case in `basic.ts` that would be great!

src/lib/types.ts
dshukertjr: @soedirgo  Okay, I will leave it as it is for now. Thanks for checking!

src/lib/PostgrestQueryBuilder.ts
dshukertjr: Thanks for checking. I will commit your suggestion.

dshukertjr:
I will work on adding the test cases now.

test/__snapshots__/index.test.ts.snap
```diff
@@ -2006,6 +2714,50 @@ Object {
...
+  "error": null,
+  "status": 200,
+  "statusText": "OK",
+}
+`;
+
+exports[`select with head:true, count:estimated 1`] = `
+Object {
+  "body": null,
+  "count": 510,
```
dshukertjr, line 2731: count: estimated outputs 510 when the actual count is 4. Does anyone have any suggestion about what to do with this?

dshukertjr:
I do have one concern about the result of `count: estimate` and `count: planned`. Other than that, I think this pull request is pretty close to be ready to be merged. Any comments, feedback is greatly welcome!

test/__snapshots__/index.test.ts.snap
dshukertjr: I just read that we need to run `ANALYZE` on the table to update the estimate count in the plan. This seems like a separate issue, so I think we should just leave this 510 for right now and maybe deal with it in a separate pull request. What does everybody think? @steve-chavez  @kiwicopple  @soedirgo

soedirgo: Oh, so `estimated` and `planned` just give estimates, so the number can fluctuate sometimes. I think it's fine to either just test for `exact` or just snapshot the `data`.

soedirgo: On second thought, not much point in testing if it doesn't test the count. Let's keep these in for now and hope it behaves...

soedirgo:
Awesome work @dshukertjr, love the extensive tests :)

Will merge this if it looks good @kiwicopple

test/__snapshots__/index.test.ts.snap
dshukertjr: Sounds good!

kiwicopple: These numbers will definitely change based on the number of tuples Postgres needs to clean up. Perhaps we can use a [Property Matcher](https://jestjs.io/docs/en/snapshot-testing#property-matchers) to check that it's returning a(ny) number?

dshukertjr: @kiwicopple Great idea. I just updated the test code to do so!

dshukertjr:
Since the last concern that I had with this PR is solved thanks to @kiwicopple , I think this PR is ready to be merged.

PR #151, opened 2021-01-28
Remove duplicate assignment of headers

kevcodez:

Tiny optimization to not set the headers twice (one call less).

If count is set, headers will be joined and set again

## What is the new behavior?

Headers will only be set once


soedirgo:
Awesome, thanks @kevcodez!

PR #152, opened 2021-02-01
Update Type Implementation (postgrest-js side)

BeeeQueue:

A **breaking** type update, which allows easier and stricter type checking. This is a combined update for this package and `supabase-js`.

supabase/supabase-js#125

- You have to pass the object you're working on into `.from<User>('users')` every time you use it.
- `.select()` does not have any intellisense help.

## What is the new behavior?

You get intellisense suggestions in `.select()`.

---

You now pass the schema into the `PostgrestClient`, and let the types handle the rest from there.

The schema type is based on what `openapi-typescript` generates, as recommended in [the supabase guide](https://supabase.io/docs/client/generating-types) (which also has to be updated as the tool has renamed).

```ts
type User = {
  username: string
  data: object | null
  age_range: string | null
  status: 'ONLINE' | 'OFFLINE'
  catchphrase: 'string' | null
}

type Schema = {
  users: User
}

const supabase = new PostgrestClient<Schema>(REST_URL)

// Incorrect table, column names will result in type errors
typedClient.from('users').select('username').eq('username', 'supabot')

// Error
typedClient.from('uesrs').select('username').eq('username', 'wooo')
typedClient.from('users').select('username').eq('not_in_table', 'wooo')
```

## Additional context

Not passing in a type still works as before, with any names working and returning `any`.

I tried very hard to not make this a breaking change, by still allowing `.from<Object>()`, but it's sadly impossible without some crazy workarounds. 😢 

This package will have to be released before `supabase-js` so #123 can be updated to use the new version.

### DIFF
--- a/src/lib/PostgrestQueryBuilder.ts
+++ b/src/lib/PostgrestQueryBuilder.ts
@@ -21,7 +21,7 @@ export default class PostgrestQueryBuilder<Type extends TableBase> extends Postg
    * @param count  Count algorithm to use to count rows in a table.
    */
   select(
-    columns: '*' | keyof Type | Array<keyof Type> = '*',
+    columns: '*' | string | keyof Type | Array<keyof Type> = '*',
     {
       head = false,
       count = null,
--- a/src/lib/PostgrestQueryBuilder.ts
+++ b/src/lib/PostgrestQueryBuilder.ts
@@ -1,13 +1,13 @@
-import { PostgrestBuilder } from './types'
+import { PostgrestBuilder, TableBase } from './types'
 import PostgrestFilterBuilder from './PostgrestFilterBuilder'
 import PostgrestTransformBuilder from './PostgrestTransformBuilder'
 
-export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
+export default class PostgrestQueryBuilder<Type extends TableBase> extends PostgrestBuilder<Type> {
   constructor(
     url: string,
-    { headers = {}, schema }: { headers?: { [key: string]: string }; schema?: string } = {}
+    { headers = {}, schema }: { headers?: Record<string, string>; schema?: string } = {}
   ) {
-    super({} as PostgrestBuilder<T>)
+    super({} as PostgrestBuilder<Type>)
     this.url = new URL(url)
     this.headers = { ...headers }
     this.schema = schema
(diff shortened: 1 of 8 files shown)


test/basic.ts
```diff
@@ -102,23 +102,75 @@ test('connection error', async () => {
...
-  // Autocomplete should show properties of user after '.'
-  user.username
+  describe('without schema', () => {
+    test('everything works without any types passed in', async () => {
+      // eq should show User's properties in LSP/IntelliSense
+      const { data: users } = await postgrest.from('users').select().eq('username', 'supabot')
+
+      // Should not error on any property
+      users[0].username
+      users[0].somethingElse
```
BeeeQueue, line 121: These are for some reason nullable, but I can't find what in the code actually does it... They weren't nullable before and they aren't in other tests..

src/lib/PostgrestQueryBuilder.ts
```diff
@@ -31,9 +31,14 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
     } = {}
   ): PostgrestFilterBuilder<T> {
     this.method = 'GET'
+
+    if (Array.isArray(columns)) {
+      columns = columns.join(',')
+    }
```
BeeeQueue, line 37: This handles the array select input

src/lib/PostgrestQueryBuilder.ts
```diff
@@ -21,7 +21,7 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
    * @param count  Count algorithm to use to count rows in a table.
    */
   select(
-    columns = '*',
+    columns: '*' | string | keyof T | Array<keyof T> = '*',
```
BeeeQueue, line 24: This will show `'*'`, and the column names as suggestions, while still allowing a comma-separated string

jonlambert:
Went digging looking for how to type `select()` calls and came across this PR 🙂 Sorry to be 'that guy' - any idea of the state of this?

kiwicopple:
We're working on this as part of the Supabase v2 release (https://github.com/supabase/supabase-js/issues/170), but that will be after Launch Week (next week). We'll do some project planning in August and should have a better idea of dates then @jonlambert

jonlambert:
Brilliant, thanks @kiwicopple. I'll keep an eye out for that release.

BeeeQueue:
Guess I can close these now.

PR #156, opened 2021-02-22
fix: rpc method

phamhieu:

- Create PostgrestRpcBuilder
- Move `rpc` method from PostgrestQueryBuilder to PostgrestRpcBuilder
- Fixes `PostgrestClient.rpc()` options param 
- Improve fetch response + error parsing (fixes https://github.com/supabase/supabase-js/issues/132)
- Fixes stored procedure unit tests

- When `rpc` method insides PostgrestQueryBuilder. User can create this chain
`let res = await postgrest.from('users').rpc({ name_param: 'supabot'})` 
It's wrong.
- no way to set options param
Screenshot 2021-02-22 at 10 14 38 AM

## What is the new behavior?

- No breaking change. `postgrest.rpc` usage is the same after moving it to PostgrestRpcBuilder
- Expose `rpc` options param
Screenshot 2021-02-23 at 9 27 46 AM

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -91,13 +91,7 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
           const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
           if (this.method !== 'HEAD' && !isReturnMinimal) {
             const text = await res.text()
-            if (text && text !== '') {
-              try {
-                data = JSON.parse(text)
-              } catch (_) {
-                error = { message: 'Failed to parse json response' }
-              }
-            }
+            if (text && text !== '') data = JSON.parse(text)
           }
 
           const countHeader = this.headers['Prefer']?.match(/count=(exact|planned|estimated)/)
@@ -106,12 +100,7 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
             count = parseInt(contentRange[1])
           }
         } else {
-          const text = await res.text()
-          try {
-            error = JSON.parse(text)
-          } catch (_) {
-            error = { message: text }
-          }
+          error = await res.json()
         }
 
         const postgrestResponse: PostgrestResponse<T> = {
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -94,7 +94,7 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
             if (text && text !== '') {
               try {
                 data = JSON.parse(text)
-              } catch (err) {
+              } catch (_) {
                 error = { message: 'Failed to parse json response' }
               }
             }
@@ -109,7 +109,7 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
           const text = await res.text()
           try {
             error = JSON.parse(text)
-          } catch (err) {
+          } catch (_) {
             error = { message: text }
           }
         }
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -83,32 +83,37 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
       body: JSON.stringify(this.body),
     })
       .then(async (res) => {
-        let error, data, count
+        let error = null
+        let data = null
+        let count = null
+
         if (res.ok) {
-          error = null
-          if (this.method !== 'HEAD') {
-            const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
-            data = isReturnMinimal ? null : await res.json()
-          } else {
-            data = null
+          const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
+          if (this.method !== 'HEAD' && !isReturnMinimal) {
+            const text = await res.text()
+            if (text && text !== '') {
+              try {
+                data = JSON.parse(text)
+              } catch (err) {
+                error = { message: 'Failed to parse json response' }
+              }
+            }
           }
 
           const countHeader = this.headers['Prefer']?.match(/count=(exact|planned|estimated)/)
-          if (countHeader) {
-            const contentRange = res.headers.get('content-range')?.split('/')
-            if (contentRange && contentRange.length > 1) {
-              count = parseInt(contentRange[1])
-            } else {
-              count = null
-            }
-          } else {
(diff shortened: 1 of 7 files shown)


phamhieu:
Are `isVoid` and `head` duplicated ? 

I'm trying to fix this issue https://github.com/supabase/supabase-js/issues/132

kiwicopple:
It looks like the function [was previously working](https://github.com/supabase/supabase/issues/762#issue-812748693). I wonder if we added something that broke it?

> add `isVoid` param

I'm guessing the library is returning an error because we are trying to do some sort of `await res.json()` on an empty result? Perhaps it's better if we we just detect `if (res.json) {} else ..`, so that the user doesn't have to add `isVoid` every time?

phamhieu:
It seems to be a postgrest error. For void function, postgrest doesn't return anything.

my test function
```sql
CREATE FUNCTION public.void_func() 
RETURNS void AS $$
$$ LANGUAGE SQL;
```

When i create a unit test. It returns properly with my PR and also with the current postgrest-js codebase.

Screenshot 2021-02-22 at 4 32 54 PM

However, when I test with a code playground with a real supabase project. postgrest doesn't complete the request for a void function. 

Screenshot 2021-02-22 at 4 29 29 PM

Screenshot 2021-02-22 at 4 29 44 PM

phamhieu:
Screenshot 2021-02-22 at 5 31 35 PM

Confirm, it's postgrest error on this version `postgrest/postgrest:nightly-2021-01-25-18-38-c93e8f9`

kiwicopple:
Let's wait to see what @steve-chavez thinks. I believe we're using the nightly version in prod so that we can update the config without downtime

soedirgo:
I see, looks like PostgREST 7.0.1 is returning `null` for void-returning functions:

```sh
$ curl -s 'http://localhost:3000/rpc/void_func' -H 'Content-Type: application/json' -H 'Accept: application/json' -d '{"param":"foo"}'
null
```

But it returns nothing on nightly.

steve-chavez:
> But it returns nothing on nightly.

Yes, this was a side-effect of fixing functions that return `SETOF` on: https://github.com/PostgREST/postgrest/commit/0c25f128250f427f470910104eea6b8bee589083.

Didn't realize it would break `JSON.parse`, I'll file it as a bug.

> isVoid: set 'Prefer': 'return=minimal' to request header. This will prevent json parsing error when calling void function
> Postgrest stored procedure rpc doesn't support HEAD request.

Postgrest do supports HEAD on RPC, but the `return=minimal` seems like a good workaround for now.

steve-chavez:
Another suggestion for fixing the issue:
```js
if (body === '') return undefined
return JSON.parse(body)
```
From https://github.com/PostgREST/postgrest/issues/1762#issuecomment-783562601.

phamhieu:
i improve fetch response parsing method. It can handle empty result

src/lib/types.ts
```diff
@@ -83,32 +83,37 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
...
+          const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
+          if (this.method !== 'HEAD' && !isReturnMinimal) {
+            const text = await res.text()
+            if (text && text !== '') {
+              try {
+                data = JSON.parse(text)
+              } catch (_) {
+                error = { message: 'Failed to parse json response' }
+              }
+            }
```
soedirgo, line 100: Maybe we should just do nothing in catch? Not sure if it's a good idea to "forge" PostgREST errors. Otherwise you'll also need other fields in PostgrestError:

https://github.com/supabase/postgrest-js/blob/9ecb9851429bb15b6569ad7dfd24b7363e335e93/src/lib/types.ts#L8-L13

Also could do without the if methinks

src/lib/types.ts
```diff
@@ -83,32 +83,37 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
...
         } else {
-          error = await res.json()
-          data = null
-          count = null
+          const text = await res.text()
+          try {
+            error = JSON.parse(text)
+          } catch (_) {
+            error = { message: text }
+          }
```
soedirgo, line 114: Ditto here.

soedirgo (changes requested):
Just a couple comments, otherwise lgtm 👍

src/lib/types.ts
phamhieu: Because we always expect json response from PostgREST. So if the response.text has content and we can't parse it to json. An error message is reasonable.

Actually, the error message is for us. I don't think users can do anything with this error. It's either PostgREST bug or postgrest-js bug.

src/lib/types.ts
phamhieu: I can revert this to the original `error = await res.json()`. 
Just want to be safe here.

src/lib/types.ts
soedirgo: Agreed, either way the user gets an unexpected error and files a bug. I guess we can keep old behavior and add special case for `text === ''`? Loud and explicit error > silent error IMO.

src/lib/types.ts
```diff
@@ -83,32 +83,37 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
...
         if (res.ok) {
-          error = null
-          if (this.method !== 'HEAD') {
-            const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
-            data = isReturnMinimal ? null : await res.json()
-          } else {
-            data = null
+          const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
+          if (this.method !== 'HEAD' && !isReturnMinimal) {
+            const text = await res.text()
```
kiwicopple, line 93: will `res.text` ever be undefined here?

kiwicopple (approved):
Nice one, thanks @phamhieu 👍

src/lib/types.ts
phamhieu: yes, https://github.com/PostgREST/postgrest/issues/1762#issuecomment-783562601

src/lib/types.ts
phamhieu: let do less. I remove try/catch block.

src/lib/types.ts
phamhieu: reverted to  error = await res.json()

soedirgo (approved):

PR #160, opened 2021-03-07
support embedded or filter

bmartel:
# Embedded filters

Postgrest supports embedded filters through the following syntax https://postgrest.org/en/v7.0.0/api.html#embedded-filters

## Use case

Filtering a sub selection in postgres with `or`/`and` for the foreignTable.

Required workaround query param format based on an extended version of the supabase todos example:

```js
const start = new Date().toISOString();
const end = new Date().toISOString();

const workaroundParam = `todos.or=(completed_at.is.null,and(completed_at.gte.${start},completed_at.lt.${end}))`

await client
.from(`projects?${workaroundParam}`)
.select('id,todos(id,completed_at)')
.eq('id', 1)
.order('inserted_at', { foreignTable: "todos", ascending: false });
```

## Solution

Update the or filter to support foreignTable handling to achieve the desired output of postgrest which doesn't require hacking it into the from statement.

```js
const start = new Date().toISOString();
const end = new Date().toISOString();

await client
.from(`projects`)
.select('id,todos(id,completed_at)')
.eq('id', 1)
.or(`completed_at.is.null,and(completed_at.gte.${start},completed_at.lt.${end})`, { foreignTable: 'todos' })
.order('inserted_at', { foreignTable: "todos", ascending: false });
```

### DIFF
--- a/src/lib/PostgrestFilterBuilder.ts
+++ b/src/lib/PostgrestFilterBuilder.ts
@@ -47,9 +47,11 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * Finds all rows satisfying at least one of the filters.
    *
    * @param filters  The filters to use, separated by commas.
+   * @param foreignTable  The foreign table to use (if `column` is a foreign column).
    */
-  or(filters: string): this {
-    this.url.searchParams.append('or', `(${filters})`)
+  or(filters: string, { foreignTable }: { foreignTable?: string } = {}): this {
+    const key = typeof foreignTable === 'undefined' ? 'or' : `${foreignTable}.or`
+    this.url.searchParams.append(key, `(${filters})`)
     return this
   }
 
--- a/test/__snapshots__/index.test.ts.snap
+++ b/test/__snapshots__/index.test.ts.snap
@@ -790,6 +790,140 @@ Object {
 }
 `;
 
+exports[`embedded filters embedded or 1`] = `
+Object {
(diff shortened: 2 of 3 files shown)


kiwicopple:
Nice one @bmartel thanks for the PR.

Will just need to check with @soedirgo - I thought this was already possible with our current [embedded filters](https://supabase.io/docs/client/filter#filter-embedded-resources) implementation. eg:

```js
  .from('projects')
  .select('id,todos(id,completed_at)')
  .or(`todos.completed_at.is.null, completed_at.gte.${start}`)
  .or(`todos.completed_at.is.null, completed_at.lt.${end}`)
```

soedirgo:
I see, single column filters like `eq`, `gte`, etc. worked fine but not for `or` & `and`, i.e. you can't do:
```
.../users?select=*,messages(*)?or=(messages.channel_id.eq.1)
```
Which gives:
```
{
  "details": "unexpected \"c\" expecting \"not\" or operator (eq, gt, ...)",
  "message": "\"failed to parse logic tree ((messages.channel_id.eq.1))\" (line 1, column 13)"
}
```
Rather you have to do:
```
.../users?select=*,messages(*)?messages.or=(channel_id.eq.1)
```

Will update for `and` as well - thanks for the PR @bmartel!

kiwicopple:
Thanks @bmartel !

@soedirgo perhaps you can bump the version in supabase-js (and add an example in the docs)?

PR #165, opened 2021-03-11
Rename filters

### LINKED ISSUE #148: Rename filters
Some of the filter names are very short. Names like `.sl()`, `.sr()`, `.adj()` can be hard to remember what to they do. I think it would lead to clearer and more declarative code if the function names were a bit longer.

Current list of filters: https://supabase.io/docs/client/using-filters

## Rename suggestions:

- `.cs()` -> `.contains()` or add a `json: true` option to `.in()`
- `.cd()` -> `.within()`
- `.sl()` -> `.rangeLt()` or add a `range` option to `.lt(column, value, {range: true})`
- `.sr()` -> `.rangeGt()` or add a `range` option to `.gt(column, value, {range: true})`
- `.nxl()` -> `rangeGte()` or add a `range` option to `.gte(column, value, {range: true})`
- `.nxr()` -> `rangeLte()` or add a `range` option to `.lte(column, value, {range: true})`
- `.adj()` -> `.adjacent()`
- `.ova()` -> `.within()`
- `.fts()` -> `.textSearch(column, search, {type: 'tsvector'})`
- `.plfts()` -> `.textSearch(column, search, {type: 'plainto'})`
- `.phfts()` -> `.textSearch(column, search, {type: 'phraseto'})`
- `.wfts()` -> `.textSearch(column, search, {type: 'websearch'})`

_Originally posted by @joshnuss in https://github.com/supabase/supabase/discussions/534_

soedirgo:

Feature.

Some filter names are very short and hard to understand.

## What is the new behavior?

Rename the filters as per https://github.com/supabase/postgrest-js/issues/148. Some deviations to the original suggestion (I'm not good at names so if these look bad feel free to chime in):
- `cd` becomes `containedBy` rather than `within` (closer to the short version)
- `ov` becomes `overlaps` rather than `within` (same here)
- `fts` becomes `type: null` (default) rather than `type: 'tsvector'` (`tsvector` is the type of the text/document rather than the query)
- `plfts` becomes `type: 'plain'` rather than `type: 'plainto'`
- `phfts` becomes `type: 'phrase'` rather than `type: 'phraseto'`

Old filter names are kept for back compat but are tagged with `@deprecated`.

## Additional context

Closes #148.


kiwicopple:
Amazing work @soedirgo 💯 . I'm curious about the `rangeGt`, `rangeGte` etc - did you choose this over `range` for a particular reason?

soedirgo:
👍

> did you choose this over range for a particular reason?

Mostly because they operate on different kinds of columns, in contrast to `textSearch` where each variant is just a different way of interpreting the `query`.

kiwicopple:
Great. I inlined some of the test snapshots so they are easier to read.

Would be interested to get @thorwebdev and  @joshnuss 's (OP) opinion on the final changes:

- `.cs()` -> `.contains()` 
- `.cd()` -> `.within()`
- `.sl()` -> `.rangeLt()` 
- `.sr()` -> `.rangeGt()`
- `.nxl()` -> `rangeGte()`
- `.nxr()` -> `rangeLte()` 
- `.adj()` -> `.adjacent()`
- `.ova()` -> `.within()`
- `.fts()` -> `.textSearch(column, search, {config: 'english'})`
- `.plfts()` -> `.textSearch(column, search, {config: 'english', type: 'plain'})`
- `.phfts()` -> `.textSearch(column, search, {config: 'english', type: 'phrase'})`
- `.wfts()` -> `.textSearch(column, search, {config: 'english', type: 'websearch'})`

@steve-chavez this look ok to you?

steve-chavez:
Looks much better!

> .adj() -> .adjacent()

Since adjancent(`-|-`) is only for range types, perhaps the filter should be named `rangeAdjacent`?

(check with `select * from pg_operator where oprname = '-|-';`)

> .overlaps()

This one looks good. overlaps(`&&`) can be applied on other types besides ranges(arrays for example).

(check with `select * from pg_operator where oprname = '&&';`)

joshnuss:
This is great! I think it will help a lot of folks. :+1:

kiwicopple:
OK great, looks like we have enough consensus here.

> Since adjancent(-|-) is only for range types, perhaps the filter should be named rangeAdjacent?

@soedirgo perhaps we rename this one as Steve suggested, then we're good to merge yeah?

soedirgo:
Yup! I've done the rename.

Issue #164, opened 2021-03-10
Handle Postgrest special characters


inian:
Postgrest defines the [some special characters which](https://postgrest.org/en/v7.0.0/api.html#reserved-characters) need to be handled separately. 

This doesn't work
```js
await postgrest
.from('table')
.select('*')
.where('name', 'file (1).jpg')
```

but double quoting the filter works. 

```js
await postgrest
.from('table')
.select('*')
.where('name', '"file (1).jpg"')
```

It would be better if the library checks if filters have special characters and automatically adds the double quotes. It seems cleaner than handling it at an application level. 
WDYT?

soedirgo:
That's odd, I tried to reproduce this with:
```sh
$ npm run test:db
$ curl -s 'http://localhost:3000/users' -H 'Content-Type: application/json' -d '{"username":"file (1).jpg"}'
```
And then:
```ts
import { PostgrestClient } from '@supabase/postgrest-js'
const postgrest = new PostgrestClient('http://localhost:3000')
postgrest
  .from('users')
  .select()
  .eq('username', 'file (1).jpg')
  .then(({ data, error }) => console.log(data, error))
// prints:
// [
//   {
//     username: 'file (1).jpg',
//     data: null,
//     age_range: null,
//     status: 'ONLINE',
//     catchphrase: null
//   }
// ] null
```
Maybe the field actually has `"` in it?

More generally re: quoting, there has been some problems when automatically escaping on special characters: https://github.com/supabase/postgrest-js/issues/131, https://github.com/supabase/postgrest-js/issues/149, though these are on column names rather than values. I suspect there are certain filters for which it's safe to automatically escape special characters for values, but I haven't found an obvious one yet. Another problem is `.eq('username', 'supabot')` is incompatible with `.eq('username', '"supabot"')`.

inian:
Field does not have `"` in it. 

`eq` works but `in` doesn't. Can you try this?
```
const body = await supabase
      .from("objects")
      .select("*")
      .in("name", ["authenticated/test-image-001 (1).png"]);
```

Also this seems to be a regression. A previous version of postgrest-js handled this without issues. I can find out which version the regression happened if it helps.

soedirgo:
I see. I remember I removed quoting on `in` on https://github.com/supabase/postgrest-js/pull/155.

I did some tests and it seems like always quoting makes it not match values that have `"` in it, i.e. if I do:
```sh
[...]

PR #166, opened 2021-03-11
fix: handle postgrest special characters in filters

inian:
fixes https://github.com/supabase/postgrest-js/issues/164

This only quotes values and not column names. Previous bugs were due to quoting column names. 

I also add in a check to only quote the value if the filter value contains a postgrest special character.

### DIFF
--- a/src/lib/PostgrestFilterBuilder.ts
+++ b/src/lib/PostgrestFilterBuilder.ts
@@ -4,16 +4,6 @@ import PostgrestTransformBuilder from './PostgrestTransformBuilder'
  * Filters
  */
 
-const cleanFilterArray = <T>(filter: T[keyof T][]) =>
-  filter
-    .map((s) => {
-      // handle postgrest reserved characters
-      // https://postgrest.org/en/v7.0.0/api.html#reserved-characters
-      if (typeof s === 'string' && new RegExp('[,()]').test(s)) return `"${s}"`
-      else return `${s}`
-    })
-    .join(',')
-
 type FilterOperator =
   | 'eq'
   | 'neq'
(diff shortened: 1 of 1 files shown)


soedirgo (approved):
LGTM 👍 though I'd wait for steve's response on properly quoting `"` itself

steve-chavez:
Something I've noticed is that the `cleanFilterArray` function is also applied to the `cs`, `cd` and `ov` operators and those don't need special handling. Their filter value is just a string, `cleanFilterArray` could be removed from them.

https://github.com/supabase/postgrest-js/blob/e5f5dfae7da22f4ff63c11c124d65d0699b7b890/test/filters.ts#L65-L68
https://github.com/supabase/postgrest-js/blob/e5f5dfae7da22f4ff63c11c124d65d0699b7b890/test/filters.ts#L70-L73
https://github.com/supabase/postgrest-js/blob/e5f5dfae7da22f4ff63c11c124d65d0699b7b890/test/filters.ts#L100-L103

soedirgo:
Gotcha, thanks Steve!

src/lib/PostgrestFilterBuilder.ts
```diff
@@ -4,7 +4,15 @@ import PostgrestTransformBuilder from './PostgrestTransformBuilder'
  * Filters
  */
 
-const cleanFilterArray = <T>(filter: T[keyof T][]) => filter.map((s) => `${s}`).join(',')
+const cleanFilterArray = <T>(filter: T[keyof T][]) =>
+  filter
+    .map((s) => {
+      // handle postgrest reserved characters
+      // https://postgrest.org/en/v7.0.0/api.html#reserved-characters
+      if (typeof s === 'string' && new RegExp('[,.():]').test(s)) return `"${s}"`
```
steve-chavez, line 12: Just `()` and `,` need quoting inside `in`. Quoting `:` and `.` is only needed when quoting column names.

```suggestion
      if (typeof s === 'string' && new RegExp('[,()]').test(s)) return `"${s}"`
```

soedirgo:
Should be good now @inian, feel free to merge 👍

inian:
Thanks Bobbie and Steve!

PR #176, opened 2021-05-08
fix: include `columns` param on bulk INSERT

### LINKED ISSUE #175: Use `columns` query parameter for insert
## Feature request

With the [columns query parameter](https://postgrest.org/en/stable/api.html#specifying-columns), the `All object keys must match` error message is avoided. Instead the direct PostgreSQL error is shown, which is more helpful for debugging.

Example:

```js

// When making a bulk insert
  .from('mytable')
  .insert([
    {
      field: 'val2',
      nonNullableField: 'val3'        
    },
    {
      field: 'val1'     
    }      
])

// Right now this happens
console.log(error);
{message: 'All object keys must match'}

// If we were to use the columns query param, we'd get a clearer error
console.log(error);
{
    "code": "23502",
    "details": "Failing row contains (val1, null).",
    "hint": null,
    "message": "null value in column \"nonNullableField\" violates not-null constraint"
}
```

Note: This doesn't solve #173 because the default value of the columns won't be applied. It would still clear the error message though.

### Possible implementation

Is not necessary to add a new method to the lib. Internally, the body of the bulk insert can be passed through a function that gets the keys and produces the `columns` query param.

Something like:

```js

let body = [
    {
      field: 'val2',
      nonNullableField: 'val3'        
    },
    {
      field: 'val1',
      anotherField: 'val4'     
    }      
];

// Using https://ramdajs.com/
let columns = R.join(',', R.uniq(R.flatten(R.map(R.keys, body))));

console.log(columns);
"field,nonNullableField,anotherField"
// append this to ?columns= in the http request
```

soedirgo:

Bug fix.

Bulk insert doesn't have [columns](https://postgrest.org/en/stable/api.html#specifying-columns) param.

## What is the new behavior?

Adds the param. Fixes #175.

### DIFF
--- a/src/lib/PostgrestQueryBuilder.ts
+++ b/src/lib/PostgrestQueryBuilder.ts
@@ -107,6 +107,12 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
 
     this.headers['Prefer'] = prefersHeaders.join(',')
 
+    if (Array.isArray(values)) {
+      const columns = values.reduce((acc, x) => acc.concat(Object.keys(x)), [] as string[])
+      const uniqueColumns = [...new Set(columns)]
+      this.url.searchParams.set('columns', uniqueColumns.join(','))
+    }
+
     return new PostgrestFilterBuilder(this)
   }
 
--- a/test/basic.ts
+++ b/test/basic.ts
@@ -234,3 +234,8 @@ describe("insert, update, delete with count: 'exact'", () => {
     expect(res).toMatchSnapshot()
   })
 })
+
+test('insert includes columns param', async () => {
+  const client = postgrest.from('users').insert([{ foo: 1 }, { bar: 2 }])
+  expect((client as any).url.searchParams.get('columns')).toMatchInlineSnapshot(`"foo,bar"`)
+})


steve-chavez (approved):
LGTM!

steve-chavez:
@soedirgo Q: How can I test this against a local postgrest? These [steps](https://github.com/supabase/postgrest-js#quick-start) don't work for me with `node`.

Is there an easy way I can try the library on the terminal?

soedirgo:
👍

> Is there an easy way I can try the library on the terminal?

This should do it:

```js
const { PostgrestClient } = require('.')

const REST_URL = 'http://localhost:3000'
const postgrest = new PostgrestClient(REST_URL)

postgrest
    .from('users')
    .select()
    .then(console.log)
```

steve-chavez:
@soedirgo Awesome. Thanks!

I've just confirmed that the PR is working.

Did this:

```js
postgrest
    .from('projects')
    .insert([
      {id: 123, name: "project-123", nonexistent: "val"},
      {id: 345, name: "project-345"}
    ])
    .then(console.log)
```

And the error is much clearer now.

```js
{
  error: {
    hint: null,
    details: null,
    code: '42703',
    message: 'column "nonexistent" of relation "projects" does not exist'
  }
  //...
}
```

PR #180, opened 2021-05-13
fix(rpc): allow filters on rpc

soedirgo:

Bug fix.

Stored procedures (`rpc()`) return `PostgrestTransformBuilder` and thus doesn't support filters (`eq()`, etc.).

## What is the new behavior?

Return `PostgrestFilterBuilder` to support filters.


kiwicopple:
fast work @soedirgo 💯

Issue #183, opened 2021-05-21
Create maybeSingle() function


kiwicopple:
We have a single() function:

```js
supabase
.from('table')
.select()
.eq('name', 'xx')
.single()
```

If this function returns either 0 rows, or >1 row, then it throws an error

## RFC

Add an option/param to this function `throws` which can have:

- `always` (default) - current behavior
- `never` - catches the error from PostgREST and just returns `{ data: null, error: null }`

this is particularly useful when there are 0 rows. For example in the latest stream: https://youtu.be/p561ogKZ63o?t=2637. In this case I dont really mind if there are no rows - if there was no data it's OK, but it's returning the error too.

The one caveat - what if the query fetches multiple rows? Then the data is `null` and it looks like "no" rows. I don't know how to circumvent this, because the error from PostgREST is the same for both 0 & >1 row.

PR #184, opened 2021-05-27
Create maybeSingle() function

kiwicopple:
- fix: #183

### DIFF
--- a/src/lib/PostgrestTransformBuilder.ts
+++ b/src/lib/PostgrestTransformBuilder.ts
@@ -1,4 +1,4 @@
-import { PostgrestBuilder, PostgrestSingleResponse } from './types'
+import { PostgrestBuilder, PostgrestMaybeSingleResponse, PostgrestSingleResponse } from './types'
 
 /**
  * Post-filters (transforms)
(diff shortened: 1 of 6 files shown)


test/transforms.ts
```diff
@@ -38,6 +38,11 @@ test('single on insert', async () => {
   await postgrest.from('users').delete().eq('username', 'foo')
 })
 
+test('maybeSingle', async () => {
+  const res = await postgrest.from('users').select().limit(0).maybeSingle()
```
steve-chavez, line 42: Oh, this test doesn't look right. Doing `/table?limit=0` is invalid in postgrest and will reply with a 416 Range Not Satisfiable: https://github.com/PostgREST/postgrest/issues/1121#issuecomment-392578452. I think this restriction can be lifted, but for now I'd suggest doing a test based on a nonexistent id `.eq('id', 123456)`.

src/lib/PostgrestTransformBuilder.ts
```diff
@@ -93,4 +93,13 @@ export default class PostgrestTransformBuilder<T> extends PostgrestBuilder<T> {
     this.headers['Accept'] = 'application/vnd.pgrst.object+json'
     return this as PromiseLike<PostgrestSingleResponse<T>>
   }
+
+  /**
+   * Retrieves only one row from the result. Result must be one row (e.g. using
+   * `limit`), otherwise this will result in an error.
+   */
```
steve-chavez, line 100: This will show in docs right? I think we should mention using `maybeSingle` on a `eq` filter instead of a `limit`. This will be the most common use case IMO, and will also avoid the previous error I mentioned.

test/transforms.ts
kiwicopple: Perfect - I was in the middle of the livestream and figured I'd need your help to figure this one out 😂

src/lib/PostgrestTransformBuilder.ts
kiwicopple: Yeah - I'll update this in the next livestream and release it

soedirgo:
Updated the implementation, feel free to review again!

steve-chavez (approved):
LGTM now!

kiwicopple:
legend, thanks Bobbie - I'll update the Quickstart docs with this new function and add it to the rest of the docs

Issue #186, opened 2021-05-30
Fetch data as CSV instead of JSON


nstrelow:
I would like to download a table or certain rows of a table in CSV format instead of JSON.

Postgres has support for this here: https://postgrest.org/en/v7.0.0/api.html#response-format

So in theory, only the Accept header needs to be made configurable + disable the parsing after getting the response.

#### Use case
I am building research tools and researchers like their dataframes build upon CSV files.

andreivreja:
> So in theory, only the Accept header needs to be made configurable + disable the parsing after getting the response.

An built in parser that converts to the same object format as it currently is would also be amazing for those of us who just want to use CSV for response size reasons. In my use case, the JSON overhead makes up for around 75% of the response size, which is pretty common when not storing large strings.

nstrelow:
Awesome your PR!
Will probably do the same for postgrest-dart, which I am actually using.

I agree, lower size using csv is probably very great, the parsing could be a bit complicated though.

soedirgo:
Closed in #187.

PR #187, opened 2021-05-30
CSV Response

andreivreja:

Feature

Response is always a JSON.

## What is the new behavior?

Ability to set response type to CSV. Usage:
```
postgrest
.from('table')
.csv()
.select()
```

## Additional context

CSV responses can be super convenient for some use cases (https://github.com/supabase/postgrest-js/issues/186) but also save plenty of resources used (wasted) by large response sizes (https://github.com/supabase/supabase-js/issues/189)

### DIFF
--- a/src/lib/PostgrestTransformBuilder.ts
+++ b/src/lib/PostgrestTransformBuilder.ts
@@ -93,4 +93,12 @@ export default class PostgrestTransformBuilder<T> extends PostgrestBuilder<T> {
     this.headers['Accept'] = 'application/vnd.pgrst.object+json'
     return this as PromiseLike<PostgrestSingleResponse<T>>
   }
+
+  /**
+   * Set the response type to CSV.
+   */
+   csv(): this {
+    this.headers['Accept'] = 'text/csv'
+    return this
+  }
 }
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -91,7 +91,7 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
           const isReturnMinimal = this.headers['Prefer']?.split(',').includes('return=minimal')
           if (this.method !== 'HEAD' && !isReturnMinimal) {
             const text = await res.text()
-            if (text && text !== '') data = JSON.parse(text)
+            if (text && text !== '' &&  this.headers['Accept'] !== 'text/csv') data = JSON.parse(text)
           }
 
           const countHeader = this.headers['Prefer']?.match(/count=(exact|planned|estimated)/)


andreivreja:
I'm thinking that a built in parser + convertor (to the same object format we're using now) would be amazing. 

What's the stance on using something like `fast-csv` or  `papaparse` for this? `papaparse` is pretty fast and wouldn't bloat the bundle size. I know you guys love using amazing open source projects, so I assume this would be fine?

kiwicopple:
Very nice PR @andreivreja 👍

inian:
The main concern with including a csv parser is the bundle size Andrei. Our bundle is [15kb minified+gzipped now](https://bundlephobia.com/result?p=@supabase/supabase-js@1.11.5) and fast-csv itself [is 13.6 kb](https://bundlephobia.com/result?p=fast-csv@4.3.6) and papaparse [is 6.5 kb](https://bundlephobia.com/result?p=papaparse@5.3.1).

andreivreja:
> The main concern with including a csv parser is the bundle size Andrei. Our bundle is [15kb minified+gzipped now](https://bundlephobia.com/result?p=@supabase/supabase-js@1.11.5) and fast-csv itself [is 13.6 kb](https://bundlephobia.com/result?p=fast-csv@4.3.6) and papaparse [is 6.5 kb](https://bundlephobia.com/result?p=papaparse@5.3.1).

Oh, I never actually checked the size of `supabase-js`, didn't expect it to be this low. Yeah, I suppose an extra 6.5kB is too high relatively speaking. 

I'll mess around with `papaparse` these days probably and see how much is actually needed, because a good chunk of what it does isn't relevant for this.

nstrelow:
Neat that this landed.

Just some nitpicking, which I came across replicating this in dart: You could also use the response header to determine if res.headers['content-type'] == 'text/csv', instead of checking the request headers.

Seems cleaner, but should still work the same in both cases I would guess.

andreivreja:
> Neat that this landed.
> 
> Just some nitpicking, which I came across replicating this in dart: You could also use the response header to determine if res.headers['content-type'] == 'text/csv', instead of checking the request headers.
> 
> Seems cleaner, but should still work the same in both cases I would guess.

Wanted to do that initially but the fact that it has `charset=utf-8` in the header (and I don't really trust that it will always have it) and that I gotta trust it's always lower case made me just go for the request header instead. Sure, could've convert the case and check if the response header contains it instead of an exact match, but I felt like that's some extra resource usage.

Might just be my micro-optimization mindset, didn't actually check if it would've made any noticeable difference - probably it wouldn't haha.

And rather than checking if `content-type` is `text/csv`, could directly check if it is `application/json` and only parse that.

nstrelow:
I see, then I will mirror your implementation, just to have it the same.

Consistency is probably better here ^^

PR #191, opened 2021-06-11
feat: Implement and export full typings

prescience-data:
- Exports all existing un-exported types (addresses https://github.com/supabase/supabase-js/issues/168) 
- Implements specific types for previously inlined types _(eg options)_
- Implements interfaces rather than concrete _("Program to an interface, not an implementation")_
- Consolidates dangling `FilterOperator` type to `types.ts`

Feature > TypeScript Types

The package does not presently export all types and contains some concrete typings without interfaces:
https://github.com/supabase/supabase-js/issues/168

## What is the new behavior?

Exports all types and adds interfaces for all previously un-typed classes.

### DIFF
--- a/src/PostgrestClient.ts
+++ b/src/PostgrestClient.ts
@@ -1,10 +1,10 @@
 import PostgrestQueryBuilder from './lib/PostgrestQueryBuilder'
 import PostgrestRpcBuilder from './lib/PostgrestRpcBuilder'
-import PostgrestFilterBuilder from './lib/PostgrestFilterBuilder'
+import { Client, ClientConfig, FilterBuilder, QueryBuilder, RpcOptions } from './lib/types'
 
-export default class PostgrestClient {
+export default class PostgrestClient implements Client {
   url: string
-  headers: { [key: string]: string }
+  headers: Record<string, string>
   schema?: string
 
   /**
@@ -14,10 +14,7 @@ export default class PostgrestClient {
    * @param headers  Custom headers.
    * @param schema  Postgres schema to switch to.
    */
-  constructor(
-    url: string,
-    { headers = {}, schema }: { headers?: { [key: string]: string }; schema?: string } = {}
-  ) {
+  constructor(url: string, { headers = {}, schema }: ClientConfig = {}) {
     this.url = url
     this.headers = headers
     this.schema = schema
@@ -28,7 +25,7 @@ export default class PostgrestClient {
    *
    * @param token  The JWT token to use.
    */
-  auth(token: string): this {
+  auth(token: string): Client {
     this.headers['Authorization'] = `Bearer ${token}`
     return this
   }
@@ -38,7 +35,7 @@ export default class PostgrestClient {
    *
    * @param table  The table name to operate on.
    */
-  from<T = any>(table: string): PostgrestQueryBuilder<T> {
+  from<T = any>(table: string): QueryBuilder<T> {
     const url = `${this.url}/${table}`
     return new PostgrestQueryBuilder<T>(url, { headers: this.headers, schema: this.schema })
   }
@@ -53,12 +50,8 @@ export default class PostgrestClient {
   rpc<T = any>(
     fn: string,
     params?: object,
-    {
-      count = null,
-    }: {
-      count?: null | 'exact' | 'planned' | 'estimated'
-    } = {}
-  ): PostgrestFilterBuilder<T> {
+    { count }: RpcOptions = { count: null }
+  ): FilterBuilder<T> {
     const url = `${this.url}/rpc/${fn}`
     return new PostgrestRpcBuilder<T>(url, {
       headers: this.headers,
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,3 +4,5 @@ import PostgrestQueryBuilder from './lib/PostgrestQueryBuilder'
 import { PostgrestBuilder } from './lib/types'
 
 export { PostgrestClient, PostgrestFilterBuilder, PostgrestQueryBuilder, PostgrestBuilder }
+
+export * from './lib/types'
--- a/src/lib/PostgrestFilterBuilder.ts
+++ b/src/lib/PostgrestFilterBuilder.ts
@@ -1,34 +1,9 @@
 import PostgrestTransformBuilder from './PostgrestTransformBuilder'
+import { FilterBuilder, FilterOperator } from './types'
 
-/**
- * Filters
- */
-
-type FilterOperator =
-  | 'eq'
-  | 'neq'
-  | 'gt'
-  | 'gte'
-  | 'lt'
-  | 'lte'
-  | 'like'
-  | 'ilike'
-  | 'is'
-  | 'in'
-  | 'cs'
-  | 'cd'
-  | 'sl'
-  | 'sr'
-  | 'nxl'
-  | 'nxr'
-  | 'adj'
-  | 'ov'
-  | 'fts'
-  | 'plfts'
-  | 'phfts'
-  | 'wfts'
-
-export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder<T> {
+export default class PostgrestFilterBuilder<T>
+  extends PostgrestTransformBuilder<T>
+  implements FilterBuilder<T> {
   /**
    * Finds all rows which doesn't satisfy the filter.
    *
@@ -36,7 +11,7 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param operator  The operator to filter with.
    * @param value  The value to filter with.
    */
-  not(column: keyof T, operator: FilterOperator, value: any): this {
+  not(column: keyof T, operator: FilterOperator, value: any): FilterBuilder<T> {
     this.url.searchParams.append(`${column}`, `not.${operator}.${value}`)
     return this
   }
@@ -47,7 +22,7 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param filters  The filters to use, separated by commas.
    * @param foreignTable  The foreign table to use (if `column` is a foreign column).
    */
-  or(filters: string, { foreignTable }: { foreignTable?: string } = {}): this {
+  or(filters: string, { foreignTable }: { foreignTable?: string } = {}): FilterBuilder<T> {
     const key = typeof foreignTable === 'undefined' ? 'or' : `${foreignTable}.or`
     this.url.searchParams.append(key, `(${filters})`)
     return this
@@ -60,7 +35,7 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param column  The column to filter on.
    * @param value  The value to filter with.
    */
-  eq(column: keyof T, value: T[keyof T]): this {
+  eq(column: keyof T, value: T[keyof T]): FilterBuilder<T> {
     this.url.searchParams.append(`${column}`, `eq.${value}`)
     return this
   }
@@ -72,7 +47,7 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param column  The column to filter on.
    * @param value  The value to filter with.
    */
-  neq(column: keyof T, value: T[keyof T]): this {
+  neq(column: keyof T, value: T[keyof T]): FilterBuilder<T> {
     this.url.searchParams.append(`${column}`, `neq.${value}`)
     return this
   }
@@ -84,7 +59,7 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param column  The column to filter on.
    * @param value  The value to filter with.
    */
-  gt(column: keyof T, value: T[keyof T]): this {
+  gt(column: keyof T, value: T[keyof T]): FilterBuilder<T> {
     this.url.searchParams.append(`${column}`, `gt.${value}`)
     return this
   }
@@ -96,7 +71,7 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param column  The column to filter on.
    * @param value  The value to filter with.
    */
-  gte(column: keyof T, value: T[keyof T]): this {
+  gte(column: keyof T, value: T[keyof T]): FilterBuilder<T> {
     this.url.searchParams.append(`${column}`, `gte.${value}`)
     return this
   }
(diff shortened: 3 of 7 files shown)


src/index.ts
```diff
@@ -4,3 +4,5 @@ import PostgrestQueryBuilder from './lib/PostgrestQueryBuilder'
 import { PostgrestBuilder } from './lib/types'
 
 export { PostgrestClient, PostgrestFilterBuilder, PostgrestQueryBuilder, PostgrestBuilder }
+
+export * from './lib/types'
```
prescience-data, line 8: Exports all types for developers to use.
Required for next PR on `supabase-js`.

src/lib/PostgrestFilterBuilder.ts
```diff
@@ -1,42 +1,17 @@
 import PostgrestTransformBuilder from './PostgrestTransformBuilder'
+import { FilterBuilder, FilterOperator } from './types'
 
-/**
```
prescience-data, line 3: Consolidates `FilterOperator` to `types.ts`

src/lib/PostgrestQueryBuilder.ts
```diff
@@ -1,11 +1,20 @@
...
 
-export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
-  constructor(
-    url: string,
-    { headers = {}, schema }: { headers?: { [key: string]: string }; schema?: string } = {}
-  ) {
+export default class PostgrestQueryBuilder<T>
+  extends PostgrestBuilder<T>
+  implements QueryBuilder<T> {
+  constructor(url: string, { headers = {}, schema }: QueryBuilderConfig = {}) {
```
prescience-data, line 17: Provides exported constructor config interface

src/lib/PostgrestRpcBuilder.ts
```diff
@@ -1,11 +1,8 @@
...
+import { FilterBuilder, PostgrestBuilder, RpcBuilder, RpcBuilderConfig, RpcOptions } from './types'
 import PostgrestFilterBuilder from './PostgrestFilterBuilder'
 
-export default class PostgrestRpcBuilder<T> extends PostgrestBuilder<T> {
-  constructor(
-    url: string,
-    { headers = {}, schema }: { headers?: { [key: string]: string }; schema?: string } = {}
-  ) {
+export default class PostgrestRpcBuilder<T> extends PostgrestBuilder<T> implements RpcBuilder<T> {
+  constructor(url: string, { headers = {}, schema }: RpcBuilderConfig = {}) {
```
prescience-data, line 5: Provides exported constructor config interface

src/lib/types.ts
```diff
@@ -1,11 +1,38 @@
...
+  | 'sr'
+  | 'nxl'
+  | 'nxr'
+  | 'adj'
+  | 'ov'
+  | 'fts'
+  | 'plfts'
+  | 'phfts'
+  | 'wfts'
+
```
prescience-data, line 29: Consolidated from `PostgresFilterBuilder.ts`

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
-export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestResponse<T>> {
+export interface Builder<T> extends PromiseLike<PostgrestResponse<T>> {
+  then<TResult1 = PostgrestResponse<T>, TResult2 = never>(
+    onfulfilled?:
+      | ((value: PostgrestResponse<T>) => TResult1 | PromiseLike<TResult1>)
+      | undefined
+      | null,
+    onrejected?: ((reason: any) => TResult2 | PromiseLike<TResult2>) | undefined | null
+  ): PromiseLike<TResult1 | TResult2>
+}
```
prescience-data, line 90: Provide a base `Builder` interface that all `Builders` implement.

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
+  select(columns?: string, options?: SelectOptions): FilterBuilder<T>
+
+  insert(values: Partial<T> | Partial<T>[], options?: InsertOptions): FilterBuilder<T>
+
+  upsert(values: Partial<T> | Partial<T>[], options?: UpsertOptions): FilterBuilder<T>
+
+  update(values: Partial<T>, options?: UpdateOptions): FilterBuilder<T>
+
+  delete(options?: DeleteOptions): FilterBuilder<T>
+}
```
prescience-data, line 139: Provide an interface for `PostgresQueryBuilder` to implement.

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
+  rangeAdjacent(column: keyof T, range: string): FilterBuilder<T>
+
+  overlaps(column: keyof T, value: string | T[keyof T][]): FilterBuilder<T>
+
+  textSearch(column: keyof T, query: string, options: TextSearchOptions): FilterBuilder<T>
+
+  filter(column: keyof T, operator: FilterOperator, value: any): FilterBuilder<T>
+
+  match(query: { [key: string]: string }): FilterBuilder<T>
+}
```
prescience-data, line 198: Provide an interface for `PostgresFilterBuilder` to implement.

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
+
+export interface RpcBuilderConfig {
+  headers?: Record<string, string>
+  schema?: string
+}
+
+export interface RpcBuilder<T> {
+  rpc(params?: object, options?: RpcOptions): FilterBuilder<T>
+}
```
prescience-data, line 211: Provide an interface for `PostgresRpcBuilder` to implement.

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
+  limit(count: number, options?: LimitOptions): TransformBuilder<T>
+
+  range(from: number, to: number, options?: RangeOptions): TransformBuilder<T>
+
+  single(): PromiseLike<PostgrestSingleResponse<T>>
+
+  maybeSingle(): PromiseLike<PostgrestMaybeSingleResponse<T>>
+
+  csv(): TransformBuilder<T>
+}
```
prescience-data, line 241: Provide an interface for `PostgresTransformBuilder` to implement.

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
+  url: string
+  headers: Record<string, string>
+  schema?: string
+
+  auth(token: string): Client
+
+  from<T = any>(table: string): QueryBuilder<T>
+
+  rpc<T = any>(fn: string, params?: object): FilterBuilder<T>
+}
```
prescience-data, line 258: Provide an interface for `PostgrestClient` to implement.

src/lib/types.ts
```diff
@@ -17,38 +44,220 @@ type PostgrestError = {
...
+
+export abstract class PostgrestBuilder<T> implements Builder<T> {
```
prescience-data, line 260: Implement the base `Builder` interface.

src/PostgrestClient.ts
```diff
@@ -14,10 +14,7 @@ export default class PostgrestClient {
    * @param headers  Custom headers.
    * @param schema  Postgres schema to switch to.
    */
-  constructor(
-    url: string,
-    { headers = {}, schema }: { headers?: { [key: string]: string }; schema?: string } = {}
-  ) {
+  constructor(url: string, { headers = {}, schema }: ClientConfig = {}) {
```
prescience-data, line 17: Provides exported constructor config interface

soedirgo:
Thanks for the PR @prescience-data! I realize you spent a lot of time on this, but I have some comments:

> Exports all existing un-exported types

I don't think unnecessarily exporting symbols is a good idea, and certainly not `export *`. The fewer exported symbols there are, the smaller the API surface, and the easier it is for the implementation to change without causing a breaking change.

> Implements specific types for previously inlined types _(eg options)_

Most of the inlined types are short (object with 2-3 properties), so I can't justify declaring types for them. Declaring types also mean you need to do 2 go-to-definitions to get the type for each of the options, and you can't document the options' TSDoc in the same place:
```js
constructor PostgrestClient(url: string, { headers, schema }?: {
    headers?: {
        [key: string]: string;
    } | undefined;
    schema?: string | undefined;
}): PostgrestClient
Creates a PostgREST client.

@param url — URL of the PostgREST endpoint.  

@param headers — Custom headers.  

@param schema — Postgres schema to switch to.
```

> Implements interfaces rather than concrete _("Program to an interface, not an implementation")_

But the builder pattern *is* the interface. The official docs, the TSDoc, the examples, the answers in discussions/issues/PRs, form the API. It's what gets expanded when we bump the semver minor version. To my understanding, programming to an interface doesn't mean every class has to implement an interface and users should only call into an "interface" programming language construct—libraries that export some functions seem more common in the JS world—but then again, I don't really worship the ground the GoF walks on.

> Consolidates dangling `FilterOperator` type to `types.ts`

Do you have a use case that needs `FilterOperator` to be exported? The main purpose of the type is for autocomplete, which you can get without importing the type. The second reason is because the union of all the operator strings is too long to display, otherwise these would've been inlined.

If you have a use case that needs the API to be expanded, that we might have overlooked (as is the case with `PostgrestResponse` and `PostgrestError`), we'll consider adding it (though it's been over a month since the issue was opened, sorry about that 🙏), but we try to minimize the API surface so the code is easy to mantain and easy to make backward-compatible.

prescience-data:
> I don't think unnecessarily exporting symbols is a good idea, and certainly not `export *`. The fewer exported symbols there are, the smaller the API surface, and the easier it is for the implementation to change without causing a breaking change.

Yes, this makes complete sense in hindsight - more just "laziness" with the `export *` on my part.

 
> Most of the inlined types are short (object with 2-3 properties), so I can't justify declaring types for them. Declaring types also mean you need to do 2 go-to-definitions to get the type for each of the options, and you can't document the options' TSDoc in the same place:


That's fair, although I do have a use case where this forces me to re-type the `options` params as 
- I declare them as an object prior to construction, and;
- My linting rules require verbose typing of all objects.

```typescript

const options: SupabaseClientOptions = {
  headers: { foo: "bar" },
  schema: env.SUPABASE_SCHEMA
}

const url: string = env.SUPABASE_URL

const key: string = env.SUPABASE_KEY

export const config: SupabaseClientConfig = {
  url,
  key,
  options
}

```

```typescript
import { config } from "./config"

const { url, key, options }: SupabaseClientConfig = config

export const supabase: SupabaseClient = createClient(
  url,
  key,
  options
)

````
 
> But the builder pattern _is_ the interface. The official docs, the TSDoc, the examples, the answers in discussions/issues/PRs, form the API. It's what gets expanded when we bump the semver minor version.

Understood - 

My use case is when I need to extend or re-build a base class exported from a library it's handy to have an abstract to implement to follow the API when it changes rather than falling out of sync without realizing it (which would be easy to do if only extending the concrete as the final class would override method signatures silently as intended). 

Having the CI fail because the new class doesn't match the interface would be my preference, but I totally see your point from the angle you outlined.

 
> Do you have a use case that needs `FilterOperator` to be exported? The main purpose of the type is for autocomplete, which you can get without importing the type. The second reason is because the union of all the operator strings is too long to display, otherwise these would've been inlined.

I use [Zod](https://github.com/colinhacks/zod) which allows a schema to be bound to any existing type, so as above, having that list provided to the schema definition would be preferable (as it would automatically fail CI when the API changes) but not essential. 

```typescript
import { z, ZodSchema } from "zod"
import { ACCOUNTS_TABLE } from "../contstants"

type FilterOperator = "eq" | "neq" | "gt" | "gte" | "lt" | "lte"

const FilterColumn: ZodSchema<keyof Account> = z.enum([
  "first_name",
  "last_name",
  "email",
  "phone",
  "status",
  "is_active"
])

const FilterOperator: ZodSchema<FilterOperator> = z.enum([
  "eq",
  "neq",
  "gt",
  "gte",
  "lt",
  "..."
])

const FilterArgsSchema = z.object({
  column: FilterColumn,
  operator: FilterOperator,
  value: z.union([z.string(), z.number(), z.boolean()])
})

export const filterAccounts = async (args: FilterArgs) => {
  try {
    const { column, operator, value } = FilterArgSchema.parse(args)
    return supabase.from<Account>(ACCOUNTS_TABLE).not(column, operator, value)
  } catch (error) {
    // Handle error
    return []
  }
}
```

> If you have a use case that needs the API to be expanded, that we might have overlooked (as is the case with `PostgrestResponse` and `PostgrestError`), we'll consider adding it (though it's been over a month since the issue was opened, sorry about that 🙏), but we try to minimize the API surface so the code is easy to mantain and easy to make backward-compatible.

Let me know if any of the above need more context? I do understand they might be edge cases though.

soedirgo:
I guess a lot of this stems from `@typescript-eslint/typedef` which you mentioned in the issue. In the README itself it says:

```
**_Note:_** requiring type annotations unnecessarily can be cumbersome to maintain and generally reduces code readability. TypeScript is often better at inferring types than easily written type annotations would allow.

**Instead of enabling `typedef`, it is generally recommended to use the `--noImplicitAny` and `--strictPropertyInitialization` compiler options to enforce type annotations only when useful.**
```

TypeScript has type inferrence and structural typing, and these are *huge* ergonomic boosts, since they render a lot of explicit typing unnecessary, unlike the old days of Java, and still maintain some degree of type soundness. To borrow from Michael Pollan: *Use types. Not too much. Mostly inferred.*

> I use [Zod](https://github.com/colinhacks/zod) which allows a schema to be bound to any existing type, so as above, having that list provided to the schema definition would be preferable (as it would automatically fail CI when the API changes)

I think that's a valid use case. But consider how the type might be misused, e.g. a library that reexports the type to their users. That's not how we expect it to be used, since a breaking change to the type necessitates a breaking change to said library.

Another problem is exhaustive checking. If you define your own `FilterOperator`, you can exhaustively check all of its variants, and we can safely add another variant without breaking that. If you use the exported type, you can only exhaustively check the variants by having a default case, and even that may be unintended.

Note that it's still possible to get the types using some of TypeScript's utility types (e.g. `Parameters<T>`) or some other ways, but we're taking a path of least resistance not to restrict it (if that's even possible), but there's no guarantee if there's breaking change (though, given [Hyrum's Law](https://www.hyrumslaw.com), what's considered a "breaking change" is in the eyes of the beholder).

But to be clear, I *do* think we should export `PostgrestResponse` and `PostgrestError` since the pros outweigh the cons—these are the last objects you get from interacting with the API, and you might want to do `if (e instanceof PostgrestError) // do stuff`.

soedirgo:
Closing this in favor of #194, but feel free to continue the discussion if you'd like :)

Issue #189, opened 2021-06-05
Add a `skipDuplicates` option to insert


steve-chavez:
Requested in https://github.com/supabase/supabase/discussions/1834. PostgREST can skip duplicates by using the `Prefer: resolution=ignore-duplicates` header as documented [here](https://postgrest.org/en/stable/api.html#upsert).

```js
const { data, error } = await supabase
  .from('cities')
  .insert(
    [
      { name: 'The Shire', country_id: 554 },
      { name: 'Rohan', country_id: 555 }
    ],
    { skipDuplicates: true })
```

The option could also be named `ignoreDuplicates`.

soedirgo:
Should this not be on `upsert()` instead? Does `on_conflict` matter with `Prefer: resolution=ignore-duplicates`?

steve-chavez:
> Does on_conflict matter with Prefer: resolution=ignore-duplicates?

Yes, it considers the `on_conflict` constraint for choosing what duplicate column to ignore.

> Should this not be on upsert() instead? 

Yeah, I think it can be included on `upsert` instead.

PR #195, opened 2021-06-22
feat: add ignoreDuplicates option to upsert

rosswaycaster:

As requested (https://github.com/supabase/supabase/discussions/1834 & https://github.com/supabase/postgrest-js/issues/189) this PR implements the `Prefer: resolution=ignore-duplicates` header to ignore duplicates in an upsert operation.

More information on the `Prefer: resolution=ignore-duplicates` header can be found here: https://postgrest.org/en/stable/api.html#upsert

Currently not implemented so the only option for an upsert is for it to be merged.

## What is the new behavior?

Now users have the option to ignore duplicates instead of merging them.


soedirgo:
👍 thanks for the PR @rosswaycaster!

PR #212, opened 2021-08-23
fix: adds X-Client-Info header

awalias:
supabase/infrastructure#1044


soedirgo:
Using `const = require` seems to break some builds: https://github.com/supabase/supabase-js/issues/241

soedirgo (approved):

PR #215, opened 2021-09-23
Allow customizing RPC request method

bnjmnt4n:

Feature.

RPC function calls are made only using `POST`.

## What is the new behavior?

Users can customize the HTTP request type to ~~`GET` or~~ `HEAD`. This allows for selecting the count of rows in a function call returning a table without selecting any data with the `HEAD` request type and `count` option.

## Implementation Notes

- I made use of `Object.entries`, but if there's any browser compatibility requirements I can just revert to a `for`-`in` loop.
  - This also required editing the `tsconfig.json` to recognize newer ES methods
- I also didn't account for the edge case where an array parameter is passed with a `GET`/`HEAD` request. Seems like there are two possible usages of this: https://postgrest.org/en/v8.0/api.html#calling-functions-with-array-parameters and https://postgrest.org/en/v8.0/api.html#calling-variadic-functions and I'm not sure behaviour which to prefer.

### DIFF
--- a/src/PostgrestClient.ts
+++ b/src/PostgrestClient.ts
@@ -49,20 +49,17 @@ export default class PostgrestClient {
    *
    * @param fn  The function name to call.
    * @param params  The parameters to pass to the function call.
+   * @param head  When set to true, no data will be returned.
    * @param count  Count algorithm to use to count rows in a table.
-   * @param method  The HTTP request method the query is sent with.
-   *                "HEAD"/"GET" queries will invoke the function in a read-only transaction.
-   *                "HEAD" can be used with `count` to get the number of rows in a table
-   *                without returning any data.
    */
   rpc<T = any>(
     fn: string,
     params?: object,
     {
-      method = 'POST',
+      head = false,
       count = null,
     }: {
-      method?: 'POST' | 'GET' | 'HEAD'
+      head?: boolean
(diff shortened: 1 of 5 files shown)


soedirgo:
Thanks for the PR! Somehow I forgot to support `HEAD` for `rpc`.

For `HEAD`, I think it's better to follow how it's done for `select` instead: https://github.com/supabase/postgrest-js/blob/5e332bf65910134885c4e34e20758164694bcdb6/src/lib/PostgrestQueryBuilder.ts#L22-L31
Looking at the docs, it doesn't seem like there's anything `GET` does that `POST` can't do, and sending params as request body is cleaner (URL has length limit and all that), so supporting `GET` probably isn't necessary.

bnjmnt4n:
One case I could think of (although certainly niche) is that service workers are unable to cache POST requests.

soedirgo:
Hmm yeah, I don't have anything against it but maybe best to leave it out until there's a feature request for it.

bnjmnt4n:
@soedirgo updated! I guess it's good to avoid unnecessary complexity until it's really required, and this does have more consistency with `select`.

soedirgo (approved):

bnjmnt4n:
Should I send a corresponding PR to integrate this into `supabase-js`?

soedirgo:
Yes please, that'd be rad 🙏

PR #216, opened 2021-09-26
feat: AbortSignal

### LINKED ISSUE #200: Aborting requests
# Feature request

## Is your feature request related to a problem? Please describe.

After the migration from Superagent to cross-fetch there doesn't seem to be a way to abort a request.
Consider this usage within React:

```ts
useEffect(() => {
  client
    .from("users")
    .select()
    .eq("id", userId)
    .then(({data}) => setUser(data[0]))

  return () => {
    // Stop in-flight request?
  }
}, [userId]);
```

## Describe the solution you'd like

Given postgrest-js is built on `cross-fetch` now, support AbortSignal.

```ts
useEffect(() => {
  const controller = new AbortController;
  client
    .from("users")
    .select()
    .eq("id", userId)
    .signal(controller.signal)
    .then(({data}) => setUser(data[0]));

  return () => {
    controller.abort();
  }
}, [userId]);
```

## Describe alternatives you've considered

This could be more generic to support other fetch options instead of strictly signal.

## Additional context

I'd be fine implementing this as a PR. Just need to know what direction the maintainers would prefer.

soedirgo:

Feature.

Closes #200.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,23 +9,15 @@ on:
 
 jobs:
   test:
-    name: Test / OS ${{ matrix.os }} / Node ${{ matrix.node }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        node: ['12']
-
-    runs-on: ${{ matrix.os }}
-
+    name: Test
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Node
-        uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
         with:
-          node-version: ${{ matrix.node }}
+          node-version: '14'
 
-      - name: Run tests
-        run: |
-          npm ci
-          npm t
+      - run: |
+          npm clean-install
+          npm run test
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,24 +8,17 @@ on:
 
 jobs:
   docs:
-    name: Publish docs / OS ${{ matrix.os }} / Node ${{ matrix.node }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        node: ['12']
-
-    runs-on: ${{ matrix.os }}
-
+    name: Publish docs
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Node
-        uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
         with:
-          node-version: ${{ matrix.node }}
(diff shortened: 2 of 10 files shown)


inian:
What about having the abortsignal as part of the [a fetch options object](https://github.com/supabase/postgrest-js/issues/200#issuecomment-926284828) instead of a chaining call? Other parameters that I can think that can fit in the same fetch config object - timeout, custom fetch implementation ([issue](https://github.com/supabase/supabase/discussions/588

soedirgo:
I'm not really a fan of exposing a third party API as part of our own, and if it's necessary (as is the case here) I'd like to minimize it as much as possible. The problem is while this is convenient *now*, it makes backward-compatibility painful to maintain:
- It increases API surface, so there's more maintenance burden
- If one day `fetch` introduces a breaking change, that necessitates a breaking change on our side (unlikely since there's a spec, but not impossible)
- If there's part of the API that we don't want (e.g. because it makes the client non-isomorphic) taking it out is a breaking change
- There's a need to vet what exactly are parts of the fetch API we're exposing, which I haven't done, but it can reveal some problems, e.g. fetch is designed for the browser, not Node.js, so there are parts of it that isn't available on Node.js without a polyfill (e.g. `AbortController`, which we only use for tests here)

inian:
I was suggesting that we come group our own options related to making the request together (timeout, number of retries, abort etc), we should defn not expose all the [fetch options](https://developer.mozilla.org/en-US/docs/Web/API/fetch#parameters) in our API.

Side note, node is implementing fetch as a global via [undici](https://github.com/nodejs/undici#undicifetchinput-init-promise) (not yet landed) and abortSignal is already implemented in node 15, so hopefully we can remove node-fetch and node-abort-controller in future versions

soedirgo:
Ah, that makes sense - I'm not sure what other options to expose though:
- timeout with fetch is usually implemented with AbortSignal using `setTimeout` and `AbortController.abort()`
- not sure if it's common for REST client libraries to support retries as part of the API, and there might be some nuances on how it's done, e.g. exponential backoff
- custom fetch implementation I think should be in the constructor (like for faunadb-js), but I think all the issues relating to it will be resolved when the upstream fix lands (which doesn't seem like it's gonna happen, honestly https://github.com/github/fetch/issues/1024)

Sorry if I'm a bit persistent on this topic 😅

> node is implementing fetch as a global via undici

Awesome, haven't heard of this - about time we ditch yet-another-HTTP-client-X and just all use fetch!

inian:
> custom fetch implementation I think should be in the constructor 

Ah you are right, these can be in the constructor. I think aws sdk also puts these options there as well.

inian (approved):

Issue #196, opened 2021-06-28
Make `filter` and `not` behave like their per-filter method counterparts


soedirgo:
# Feature request

## Is your feature request related to a problem? Please describe.

Right now `filter` and `not` just takes the operator and value as a string without additional preprocessing. This means e.g. we can't use lists in `NOT IN` (see https://github.com/supabase/supabase/discussions/2055). It also causes some confusion when it comes to filter methods that use a different name to PostgREST, e.g. `cs` vs. `contains`.

## Describe the solution you'd like

- Make `not` switch on the operator and basically copy the code for each filter methods,
- make `not` accept e.g. `contains`, `containedBy`, etc.,
- keep `filter` as-is, but repurpose it as some sort of escape hatch.

This allows us to make per-filter methods and `not` cover 99% of the cases, and for the 1% we can use `filter`.

Question: what do we do about options like `foreignTable`, textSearch `config` in `not`?

## Describe alternatives you've considered

None.

## Additional context

https://github.com/supabase/supabase/discussions/2055
https://github.com/supabase/postgrest-dart/issues/32

soedirgo:
As per https://github.com/supabase/postgrest-js/pull/279#issuecomment-1147334327, we want to eventually move away from stringly-typed params.

The current way these filters work, which requires raw PostgREST syntax, is awkward, but I doubt this proposal would make it that much better, so closing this.

PR #225, opened 2021-11-16
Fix/clean array value on not filter

mahdiarn:

Bug fix on passing array value on `not` filter with `in` operator
Add another test case regarding the fix

throw an error when we passed array on value param of not filter. e.g. `.not('column', 'in', [1,2,3])`
```
error - unhandledRejection: {
  message: '"failed to parse filter (not.in.)" (line 1, column 8)',
  details: 'unexpected end of input expecting "("'
}
```
Related issue https://github.com/supabase/postgrest-js/issues/196#issue-931130462

## What is the new behavior?

The new behaviour provide us on calling `not` and `in` with array as value

## Additional context

none

### DIFF
--- a/src/lib/PostgrestFilterBuilder.ts
+++ b/src/lib/PostgrestFilterBuilder.ts
@@ -59,7 +59,21 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param value  The value to filter with.
    */
   not(column: keyof T, operator: FilterOperator, value: any): this {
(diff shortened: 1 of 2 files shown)


soedirgo:
Sorry for the lack of response on this! This looks like it would affect not only `not.in`, but also `not.cs`, `not.cd`, etc.

As per https://github.com/supabase/postgrest-js/pull/279#issuecomment-1147334327, we want to move away from stringly-typed params eventually (ideal DX is people shouldn't need to escape anything), so closing this.

PR #226, opened 2021-11-18
Adding notIn method as filter

pebosi:
Currently it's not possible to call "not.in" operator with an array, so i did a small copy of the in method.

Just, saw #225, its another approach.


soedirgo:
We want to avoid having redundant filters, so closing this. Hopefully in the next DX redesign you won't need to cleanse anything and can pass in arrays, plain objects, etc.

PR #242, opened 2022-01-11
fix: Fetch method can't handle error response that is not a valid JSON

### LINKED ISSUE #241: Fetch method can't  handle error response that is not a valid JSON object
# Bug report

## Describe the bug

related to https://github.com/supabase/supabase/issues/4464

https://github.com/supabase/postgrest-js/blob/ff790caf45cf3a3d8fe84e743600393988fc7122/src/lib/types.ts#L126-L132

Fetch method tries to parse error response as json before returning. Whenever the error response is not JSON object, it will not return properly instead a json parsing error will be throw.

## Expected behavior

- get error response as raw text
- trying to parse error response as json
  - if succeed: return as json
  - else: return as text

BigOsvaap:

Bug fix.

The fetch method tries to parse error response as JSON before returning. Whenever the error response is not a JSON object, it will not return properly instead a JSON parsing error will be thrown.

## What is the new behavior?

The fetch method tries to parse error response as JSON before returning, in case of failure returns a raw text error.

## Additional context

Fix #241

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -124,14 +124,18 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
           count = parseInt(contentRange[1])
         }
       } else {
+        const body = await res.text()
+
         try {
-          error = await res.json()
-        } catch (e) {
-          error = await res.text()
-        } finally {
-          if (error && this.shouldThrowOnError) {
-            throw error
+          error = JSON.parse(body)
+        } catch {
+          error = {
+            message: body
           }
+        }
+
+        if (error && this.shouldThrowOnError) {
+          throw error
         }
       }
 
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -124,10 +124,14 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
...
-        if (error && this.shouldThrowOnError) {
-          throw error
+        try {
+          error = await res.json()
+        } catch (e) {
+          error = await res.text()
+        } finally {
+          if (error && this.shouldThrowOnError) {
+            throw error
+          }
```
soedirgo, line 134: We can't reuse the response body since it was consumed with `res.json()`. Also, we want the error to be an object with a `message` property. So how about:

```js
        const body = await res.text()

        try {
          error = JSON.parse(body)
        } catch {
          error = {
            message: body
          }
        }

        if (error && this.shouldThrowOnError) {
          throw error
        }
```

soedirgo (changes requested):
Thanks for the PR @BigOsvaap! Left a comment below.

soedirgo:
Hey @BigOsvaap, just realized we urgently need the fix so I'll merge this + my suggestion in. Thanks again for the PR!

soedirgo (approved):

Issue #243, opened 2022-01-15
Support to add `Prefer` headers


bhvngt:
# Feature request
I was trying to use [Prefer: tx=rollback](https://postgrest.org/en/stable/configuration.html#db-tx-end) header as part of my regular testing. I was using following code snippet to initialise postgrest client

```
createClient(url, key, { options: { headers: { Prefer: "tx=rollback" } } } )
```

However, I realised that it gets overwritten by the current [implementation](https://github.com/supabase/postgrest-js/blob/eab227d2fb0eacd5fc4b1bc355183050a5ca40d9/src/lib/PostgrestQueryBuilder.ts#L112). 

## Is your feature request related to a problem? Please describe.
I would like to use [Prefer: tx=rollback](https://postgrest.org/en/stable/configuration.html#db-tx-end) as part of my test suite. With the current implementation I am not able to add this headers to postgrest client.

## Describe the solution you'd like
Ability to add `Prefer` headers while creating postgrest client

PR #244, opened 2022-01-15
feat: support for custom prefer headers #243

bhvngt:
Adds the ability to add custom `prefer` headers such as [Prefer: tx=rollback](https://postgrest.org/en/stable/configuration.html#db-tx-end). 

Custom prefer headers will take precedence over the existing `prefer` heaaders.

Bug fix, feature, docs update, ...
#243 

Current [implementation](https://github.com/supabase/postgrest-js/blob/eab227d2fb0eacd5fc4b1bc355183050a5ca40d9/src/lib/PostgrestQueryBuilder.ts#L112) overwrites custom `Prefer` headers that are passed during client initialisation

## What is the new behavior?

Custom Prefer Headers gets prepended to the existing prefer headers.

### DIFF
--- a/src/lib/PostgrestQueryBuilder.ts
+++ b/src/lib/PostgrestQueryBuilder.ts
@@ -108,7 +108,9 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
     if (count) {
       prefersHeaders.push(`count=${count}`)
     }
-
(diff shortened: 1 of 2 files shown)


soedirgo:
Ah, missed this one - thanks!

soedirgo (approved):

PR #247, opened 2022-02-25
fix: illegal invocation on fetch

soedirgo:
Because JS is such a beautiful language.

`globalThis` is neater but might hurt compatibility (2019+)


inian (approved):
Hahaha

PR #266, opened 2022-04-12
feat: add toString() method to PostgrestBuilder

steinroe:

Feature

Currently calling toString returns `[object Object]` and is thereby not useable as e.g. a cache key in swr, see https://github.com/vercel/swr/issues/1846 for a related issue.

## What is the new behavior?

This PR implements the `toString()` method on the `PostgrestBuilder` that returns a stringified copy of the URL with sorted search params and is thereby independent of the order at which the filters are applied.

## Additional context

This PR does not sort within the filters, e.g. 

```ts
  const string1 = postgrest.from('users').select('*').or('first_name.eq.Peter,username.eq.missing').eq('username', 'missing').toString()
  const string2 = postgrest.from('users').select('*').eq('username', 'missing').or('username.eq.missing,first_name.eq.Peter').toString()
```

would still return different strings. This can be controlled on user-side though and would add unnecessary complexity to the client.

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -75,6 +75,15 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
   }
 
   /**
+   * Returns a stringified copy of the URL with sorted search params
+   */
+  toString() {
+    const copiedURL = new URL(this.url.toString())
+    copiedURL.searchParams.sort()
+    return copiedURL.toString()
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


soedirgo:
I don't think a `toString()` method belongs in the builder class. I'm not familiar with SWR, but maybe the solution here is to create a wrapper class `W` containing `W.supabase` that returns `W.supabase.url` when `toString()` is called.

steinroe:
Thanks for your review and suggestion! I will look into it.

PR #267, opened 2022-04-17
wip: automatic retries for transient failures

darora:


test/server.ts
```diff
@@ -0,0 +1,61 @@
+import fastify from 'fastify'
+
+function build(opts = {}) {
+  const app = fastify(opts)
+  let count_503 = 1, count_500 = 1, count_400 = 1, count_524 = 1, count_504 = 1
+
+  app.get('/temporary503failure', async function(request, reply) {
+    if (count_503 < 2) {
+      count_503++;
```
soedirgo, line 9: `npm run format`

src/lib/types.ts
```diff
@@ -50,6 +50,12 @@ export type PostgrestSingleResponse<T> =
   | PostgrestResponseFailure
 export type PostgrestMaybeSingleResponse<T> = PostgrestSingleResponse<T | null>
 
+const delay = (ms: number) => {
+  return new Promise((resolve) => setTimeout(() => resolve(), ms))
+}
+
+const RetryableErrorCodes = new Set([503, 504, 522, 523, 524])
```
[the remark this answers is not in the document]
steve-chavez: I think 504, 522, 523, 524 are fine to retry on.

503 means that PostgreSQL is unreachable for postgrest, retrying in that case could be handled in another PR since PostgREST provides a Retry-After header. Later on 503 will also signal when postgrest's pool is overloaded.

src/lib/types.ts
```diff
@@ -88,6 +94,35 @@ export abstract class PostgrestBuilder<T> implements PromiseLike<PostgrestRespon
     return this
   }
 
+  retryingFetch(url: string, opts: any, retries: number, retryDelayMs: number): Promise<Response> {
+    // no automatic retries for non-read methods
+    if (!['GET', 'HEAD'].includes(this.method)) {
+      return Promise.resolve(this.fetch(url, opts))
+    }
```
steve-chavez, line 101: This is a good call. How should we handle PATCH requests that time out as on https://github.com/supabase/supabase/issues/6813?

steve-chavez: @darora Perhaps for this type of cases, we can increase the Cloudflare timeout?

https://support.cloudflare.com/hc/en-us/articles/115003011431-Troubleshooting-Cloudflare-5XX-errors#524error

> Enterprise customers can increase the 524 timeout up to 6000 seconds using the [proxy_read_timeout API endpoint](https://api.cloudflare.com/#zone-settings-change-proxy-read-timeout-setting).

soedirgo: If PostgREST treats PATCH (and DELETE) as semantically idempotent it should be fine, but maybe safer to do it only for reads for now just to limit the impact, and note it as a TODO.

steve-chavez: > If PostgREST treats PATCH (and DELETE) as semantically idempotent it should be fin

@soedirgo They are idempotent if we stick to the definition in [rfc7231](https://www.rfc-editor.org/rfc/rfc7231#section-4.2.2):

> A request method is considered "idempotent" if the intended EFFECT ON THE SERVER of multiple identical requests with that method is the same as the effect for a single such request.
> It knows that repeating the request will have the same intended effect, even if the original request succeeded, though the response might differ.

The last paragraph applies(response might be different) when using `Prefer: return=representation`, so a DELETE would return contents the first time but an empty array the next time is tried on the same resource. Not sure if would be safe to retry in those cases, as it could cause application errors?

The way around that, would be not forcing the `Prefer: return=representation`(which IIRC, is also needed for RLS policies to work as intended without `return=minimal`) on `postgrest-js`, and then retry only for cases which do not specify that header.

soedirgo: > retry only for cases which do not specify that header

Yup sounds good

> PATCH should be safe to retry on, even with Prefer: return=representation

Wouldn't it return different rows? I'm assuming return=representation returns the old rows

soedirgo: I guess it still may not be idempotent even with `return=minimal`, because you might have a trigger that itself isn't idempotent (extreme example: on DELETE charge a Stripe payment w/ pg_net)

steve-chavez: > Wouldn't it return different rows? I'm assuming return=representation returns the old rows

Hm, `return=representation` returns the new updated row. Or do you mean the row would have been modified by another PATCH request?

soedirgo: > return=representation returns the new updated row

Ah I see, you can ignore that part

steve-chavez:
An update here, PostgREST main branch has a `db-pool-acquisition-timeout` config which can be configured to convert all Cloudflare or other proxies timeout errors to a PostgREST error. Right now it responds with a generic `504 Gateway Timeout` but I'm thinking of chaging it to a `529 Site is overloaded`.

References:

- https://github.com/PostgREST/postgrest/issues/2348
- https://github.com/PostgREST/postgrest/pull/2449

PR #275, opened 2022-05-31
feat: obsolete deprecated stuff

soedirgo:
Please review https://github.com/supabase/postgrest-js/pull/274 first 🙏

BREAKING CHANGE: obselete the following previously-deprecated features:
- filter shorthands (e.g. `cs` vs. `contains`)
- `body` in response (vs. `data`)
- `upsert`ing thru `.insert()`

PR #276, opened 2022-05-31
feat: set `Prefer: return=minimal` by default

soedirgo:
Please review https://github.com/supabase/postgrest-js/pull/275 first 🙏

BREAKING CHANGE: set `Prefer: return=minimal` by default everywhere

We previously set `returning = 'representation'` by default so that
inserted/updated/deleted rows are returned. We now change the default so that
inserted/updated/deleted rows are NOT returned.

To return inserted/updated/deleted rows, call `.select()` at the end of
the call chain.

### DIFF
--- a/src/PostgrestQueryBuilder.ts
+++ b/src/PostgrestQueryBuilder.ts
@@ -74,22 +74,19 @@ export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
    * Performs an INSERT into the table.
    *
    * @param values  The values to insert.
-   * @param returning  By default the new record is returned. Set this to 'minimal' if you don't need this value.
    * @param count  Count algorithm to use to count rows in a table.
    */
   insert(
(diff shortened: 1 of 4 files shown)


steve-chavez:
Maybe one way to improve DX could be to assume `return=representation` when doing `select()`. So instead of doing:
```js
  const res = await postgrest
    .from('users')
    .upsert({ username: 'dragarcia' }, { onConflict: 'username', returning: 'representation' })
```
The user would do:
```js
  const res = await postgrest
    .from('users')
    .upsert({ username: 'dragarcia' }, { onConflict: 'username'})
    .select('*')
```

soedirgo:
I agree that does look cleaner. Will update the PR!

steve-chavez (approved):

kiwicopple (approved):
yup, I like it with the `.select()` chaining

Issue #239, opened 2022-01-06
Supabase default of nullsLast can cause performance issues with ascending:false.


GaryAustin1:
# Bug report

## Describe the bug
I'm entering this as a bug as it is not expected default behavior, but it could be feature request or documentation enhancement.

There can be a performance issue for large tables with an index and using limit when ordered by descending because of default choice enforced by postgrest.js.
  
In Postgres  NULLS FIRST is the default for ORDER BY DESC and NULLS LAST is default for the default ORDER BY ASC.

In PostgREST nullsfirst or nullslast is an optional parameter and would allow Postgres to sort by default order.

Supabase postgrest.js defaults to nullslast and can only be over ridden with a nullsFirst:true option.

In the case of using ascending:false option in Supabase.js (postgrest.js), nullslast is sent to postgREST by default and then Postgres will not use the default NULLS FIRST.

Many users will not even be thinking of nullsfirst or nullslast as default normally works fine.

## To Reproduce
The issue is discussed and "resolved" in this thread on discord:

https://discord.com/channels/839993398554656828/928590123011031050

## Expected behavior

Either it should be documented that it is recommended to also set nullsFirst:true when using ascending:false and why
OR better postgrest.js should take optional nullsFirst AND nullsLast parameter so default is not specifying and letting PostgREST and Postgres make their default choices.

## Screenshots

![image](https://user-images.githubusercontent.com/54564956/148399564-c3f4fe8e-3590-4463-8ea6-0eba9f9ab3a2.png)

## System information

Lastest Supabase.

steve-chavez:
@kiwicopple I think postgrest-js should not default to a NULLS LAST, otherwise the intuitive way of creating an index to speed up order won't work:

```sql
CREATE INDEX index_name ON TABLE table_name (date)
-- they'd have to create it as
-- CREATE INDEX index_name ON TABLE table_name (date DESC NULLS LAST)
```

> better postgrest.js should take optional nullsFirst AND nullsLast parameter so default is not specifying and letting PostgREST and Postgres make their default choices.

Agree, this would be a breaking change in postgrest-js though. Something for supabase-js 2.0 https://github.com/supabase/supabase-js/issues/170.

Some references:

- https://stackoverflow.com/questions/45719419/how-to-create-an-index-properly-for-desc-nulls-last-sorting
- https://dba.stackexchange.com/questions/254731/why-does-order-by-nulls-last-affect-the-query-plan-on-a-primary-key

[...]

PR #277, opened 2022-05-31
perf: nullsFirst = true by default

soedirgo:
BREAKING CHANGE: nullsFirst = true by default

We previously set `nullsFirst = false` by default, but it caused some perf issues: https://github.com/supabase/postgrest-js/issues/239

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -46,7 +46,7 @@ export default class PostgrestTransformBuilder<T> extends PostgrestBuilder<T> {
     column: keyof T,
     {
       ascending = true,
-      nullsFirst = false,
+      nullsFirst = true,
       foreignTable,
     }: { ascending?: boolean; nullsFirst?: boolean; foreignTable?: string } = {}
   ): this {


soedirgo:
@steve-chavez before I continue with this, is there a difference in behavior for PostgREST between `nullsfirst` and not specifying `nullsfirst`/`nullslast`? Does PostgREST infer the default based on index etc.?

steve-chavez:
> Does PostgREST infer the default based on index etc.?

No, it just doesn't specify a `NULLS x`, making it behave like PostgreSQL default:

> By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.

From https://www.postgresql.org/docs/current/queries-order.html

steve-chavez (approved):

kiwicopple (approved):

GaryAustin1:
Just want to double check on this.  My understanding is default operation of Postgres you don't want to specify anything for the nulls x parameter.  Then Postgres will pick the best way.  If you do specify, you need to look at sort by direction to decide which way.  I think you just reversed the issue.  

Is not better to just have both nullsfirst and nullslast as options with nothing as the default?

soedirgo:
I think you're right - reading the docs again, it's `NULLS LAST` by default for `DESC` and `NULLS FIRST` by default for `ASC`. I'll make another change for this.

PR #278, opened 2022-05-31
feat: use undefined as bottom value instead of null

soedirgo:
BREAKING CHANGE: use undefined as bottom value instead of null


steve-chavez:
Why is this needed?

soedirgo:
Here's the discussion on this: https://github.com/supabase/supabase-js/issues/244. Basically it doesn't make sense to treat `null` as a sentinel value when `undefined` is already implied for optional fields. E.g. `null` is redundant in `count?: null | 'exact' | ...` because it also accepts `undefined` anyway.

Note that this change only applies to function args etc. - it doesn't modify PostgREST's response at all. That way it's more consistent in that absence of value in the client lib is `undefined` while `null` is only ever used to represent a Postgres `null`, e.g. for the `.is()` filter.

steve-chavez (approved):

Issue #217, opened 2021-10-02
Idea: parse `select` queries to give better types


bnjmnt4n:
# Feature request/Potential idea

## Is your feature request related to a problem? Please describe.

When selecting columns to return from a request, if there's any complex select query with specific columns selected and resource embedding, you would need to manually specify the return types if you'd like good TypeScript checking. This can lead to a lot of duplication between the columns you're selecting, and the types you specify.

Example:
```ts
type WorkspaceWithTeam =
  definition["workspaces"] & {
    team: {
      user: Pick<definition["users"], "id" | "email">;
    };
  };

const { data, error } = await supabase
  .from<WorkspaceWithTeam>("workspaces")
  .select(`
    *,
    team:members(
      user:users(
        id,
        email
      )
    )
  `);
```

## Describe the solution you'd like

I've been experimenting with TypeScript template literals, and managed to implement a parser for simple PostgREST `select` queries: https://github.com/bnjmnt4n/postgrest-query. The project is still incomplete, requires a bit of a rewrite and should only be used with TypeScript 4.5+ due to some complexities with template literals, but I wanted to see if there's any interest in integration with `postgrest-js`.

Here's a screenshot of the LSP type:

![image](https://user-images.githubusercontent.com/813865/135711411-6e4368e3-99be-42bb-b652-3332c702a7c4.png)

The types can parse the query to see which columns are being selected, as well as which resources are being embedded. Then, assuming `definitions` is a `Record<TableName, Record<ColumnName, ColumnType>`, it can construct the expected output object type.

After integration of something like https://github.com/supabase/postgrest-js/issues/172, here's a potential implementation of a separate type export of a `TypedSupabaseClient` which would be had would be opt-in:

```ts
const supabase = createTypedClient<definitions>(URL, API_KEY);

const { data, error } = await supabase
  .from("workspaces")
  .select(`
    *,
    team:members(
      user:users(
        id,
        email
      )
    )
  `);

// data is properly typed.
```

## Describe alternatives you've considered

I think this has the potential to be useful, but integrating into `postgrest-js` might also not be required, and I can probably maintain this externally.

soedirgo:
This looks rad - would be great if this could be made to work with type generation: https://github.com/supabase/postgres-meta/issues/103.

I think the hairy part of implementing this is on the less exercised nooks and crannies of the PostgREST syntax:
- https://postgrest.org/en/v8.0/api.html#computed-cols
[...]

Issue #172, opened 2021-03-08
[Feature request] Improve syntax for using typescript "from" function


Nick-Mazuk:
# Feature request

First off, it's great that I can use TypeScript to statically type check everything. I think the syntax could be improved, though.

Right now, you need to add the definitions to every database operation:

```ts
supabase.from<definitions['table']>('table').select('*')
```

This works, but isn't optimal for a few reasons:

- Need to declare the type for each `from` call
- Need to declare the table name multiple times (prone to mistyping the table name)
- The input to `from` is not checked (e.g., you could time `from('tables')` for the above snippet and it would not fail linting)

But what if the definitions were added to the supabase client instead?

```ts
// lib file to create the client
import { createClient } from '@supabase/supabase-js'

export const supabase = createClient<definitions>(supabaseUrl, supabaseKey)

// file where you use the client
supabase.from('table').select('*')
```

Then, you could be confident in your static type checking:

- No need to duplicate the table name in the `from` function
- It's guaranteed that all database calls are type checked
- The `from` input could be type-checked (i.e., it's just the `keyof definitions`)
- Cleaner code syntax all around
- Added benefit that in VS Code, you get autocomplete on the table names, making the code more self-documenting

baba43:
This is still open, right? 
Because I thought about exacly the same.

soedirgo:
This is resolved in https://github.com/supabase/postgrest-js/pull/279. It's still a prerelease right now, but will be released as v2 in a few weeks.

Issue #265, opened 2022-04-10
Add ability to pass typed array to `select()` method


prescience-data:
# Feature request

Currently, the `select()` method of the `QueryBuilder` accepts a string of comma-separated values.

https://github.com/supabase/postgrest-js/blob/v0.37.2/src/lib/PostgrestQueryBuilder.ts#L33

This does not allow for any type-checking and feels a bit "raw" (personal opinion only).

## Describe the solution you'd like

Allow a typed array to be passed as the `columns` property.

```typescript
export default class PostgrestQueryBuilder<T> extends PostgrestBuilder<T> {
// ... 
  select (
    columns: (keyof T )[] | string = '*',
    {
      head = false,
      count = null,
    }: {
      head?: boolean
      count?: null | 'exact' | 'planned' | 'estimated'
    } = {}
  ): PostgrestFilterBuilder<T> {
    this.method = 'GET'
    // Remove whitespaces except when quoted
    let quoted = false
    
    const cleanedColumns = (Array.isArray(columns) ? columns.join(',') : columns)
      .split('')
      .map((c) => {
        if (/\s/.test(c) && !quoted) {
          return ''
        }
        if (c === '"') {
          quoted = !quoted
        }
        return c
      })
      .join('')
    // ...
  }
}
```

## Describe alternatives you've considered

This is more a "nice to have" as it works fine as is, and this approach can be implemented manually. 

However, given the minimal overhead and typing upside, I thought it might be worthwhile suggesting.

PR #279, opened 2022-06-02
Improved typings

soedirgo:
BREAKING CHANGE: Rework typings.

Closes https://github.com/supabase/postgrest-js/issues/172, closes https://github.com/supabase/postgrest-js/issues/217.

The current typings DX is not great. The way users supply types is through `.from<T>()` which needs to be done on each call. What ends up happening most of the time is users would leave the `T` which defaults to `any`, because this is the path of least resistance.

Another sticking point is the `T` above also becomes the query result type, even though you can e.g. specify only certain columns.

This change intends to alleviate the above by 1) accepting type definitions at the constructor, and 2) parsing `.select()` query to infer the result type.

As an example, using the type definitions in `test/types.ts`:

https://github.com/supabase/postgrest-js/blob/92beeddcab1ca21de493d604819376315a15d6b6/test/types.ts#L1-L124

We can use it like:

```ts
import { Database } from './types'

const postgrest = new PostgrestClient<Database>('https://...')

const res = await postgrest
  .from(
    'users' // 'users' | 'channels' | 'messages'
  ).update(
    { data: 'foo' }, // {
                     //   username?: string
                     //   data?: Json | null
                     //   age_range?: string | null
                     //   status?: 'ONLINE' | 'OFFLINE' | null
                     //   catchphrase?: string | null
                     // } | { ... }[]
  ).eq(
    'data', // 'username' | 'data' | 'age_range' | 'status' | 'catchphrase'
    'foo',   // string | null
  ).select('name:username, status') // <- renames work too

const {
  data,  // {
         //   name: string  <- renamed
         //   status: string
         // }[] | undefined
  error, // PostgrestError | undefined
} = res
```

Most of the hard work in `.select()` query parsing is done by @bnjmnt4n in https://github.com/bnjmnt4n/postgrest-query. Here we're only implementing a subset of it because certain things are tricky to do properly, e.g.:
- column casting (`::`)
- json paths (`->`, `->>`)
- embedded tables in select query parsing (`members(id)`)
- embedded column types in filters (`members.id`)

Embedded tables in `.select()` is especially tricky because it requires relation cardinality information, which can be calculated using FK information in the DB. e.g. for `.from('orgs').select('members(*)')` `orgs -> members` can be 1-to-many so `members` would be an array, but for the opposite i.e. `.from('members').select('orgs(*)')` `members` would be `object | null`. Right now all embedded tables are typed as `unknown`.

### DIFF
--- a/src/PostgrestClient.ts
+++ b/src/PostgrestClient.ts
@@ -2,12 +2,20 @@ import PostgrestQueryBuilder from './PostgrestQueryBuilder'
 import PostgrestFilterBuilder from './PostgrestFilterBuilder'
 import PostgrestBuilder from './PostgrestBuilder'
 import { DEFAULT_HEADERS } from './constants'
-import { Fetch } from './types'
+import { Fetch, GenericSchema } from './types'
 
-export default class PostgrestClient {
+export default class PostgrestClient<
+  Database = any,
+  SchemaName extends string & keyof Database = 'public' extends keyof Database
+    ? 'public'
+    : string & keyof Database,
+  Schema extends GenericSchema = Database[SchemaName] extends GenericSchema
+    ? Database[SchemaName]
+    : any
+> {
   url: string
   headers: Record<string, string>
-  schema?: string
+  schema?: SchemaName
   fetch?: Fetch
   shouldThrowOnError: boolean
 
@@ -27,7 +35,7 @@ export default class PostgrestClient {
       throwOnError = false,
     }: {
       headers?: Record<string, string>
-      schema?: string
+      schema?: SchemaName
       fetch?: Fetch
       throwOnError?: boolean
     } = {}
@@ -56,9 +64,12 @@ export default class PostgrestClient {
    *
    * @param table  The table name to operate on.
    */
-  from<T = any>(table: string): PostgrestQueryBuilder<T> {
-    const url = `${this.url}/${table}`
-    return new PostgrestQueryBuilder<T>(url, {
+  from<
+    TableName extends string & keyof Schema['Tables'],
+    Table extends Schema['Tables'][TableName]
+  >(table: TableName): PostgrestQueryBuilder<Table> {
+    const url = new URL(`${this.url}/${table}`)
+    return new PostgrestQueryBuilder<Table>(url, {
       headers: { ...this.headers },
       schema: this.schema,
       fetch: this.fetch,
@@ -70,14 +81,17 @@ export default class PostgrestClient {
    * Perform a function call.
    *
    * @param fn  The function name to call.
-   * @param params  The parameters to pass to the function call.
+   * @param args  The parameters to pass to the function call.
    * @param options  Named parameters.
    * @param options.head  When set to true, no data will be returned.
    * @param options.count  Count algorithm to use to count rows in a table.
    */
-  rpc<T = any>(
-    fn: string,
-    params: Record<string, unknown> = {},
+  rpc<
+    FunctionName extends string & keyof Schema['Functions'],
+    Function_ extends Schema['Functions'][FunctionName]
+  >(
+    fn: FunctionName,
+    args: Function_['Args'] = {},
     {
       head = false,
       count,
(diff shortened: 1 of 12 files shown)


src/PostgrestFilterBuilder.ts
```diff
@@ -60,8 +67,8 @@ export default class PostgrestFilterBuilder<T> extends PostgrestTransformBuilder
    * @param column  The column to filter on.
    * @param value  The value to filter with.
    */
-  eq(column: keyof T, value: T[keyof T]): this {
-    this.url.searchParams.append(`${column}`, `eq.${value}`)
+  eq<ColumnName extends string & keyof Table>(column: ColumnName, value: Table[ColumnName]): this {
```
alaister, line 70: Looking super cool @soedirgo!

Just wondering if there are plans to make this `ColumnName` type even more helpful?
I see you're now allowing all strings plus suggesting the keys of the table – is it also possible to suggest the keys for embedded tables when the user provides the type?

[For example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyhCGwoF4oG8BQUoEsAmAXFAM7ABOOAdgOZZRkD2A7kaRTXZQK4C2R3PAEYQydHgwBuOCAH18rclVoBfDBlCQoAWUnSU6OvJKKO2YDmAAbCAva1sxBMGJE4iANoBdOsNIzHiK5OGKrq4NAA0hAgxADyAGYAPAAqADRQUSBQEAAewBCUeMRQANbRDPFQyfplIBVVAHz6mdl5BUXGdnTYAPxV7pmerfmFxQAUVPEiUACqAJRe3dhQfQAGACRomcoAdJuZcUkzDcqrS9hEyQPRQ7kjHQBKEADGDGR4iWxK6fCUIA3nZZrfbRXYgmIJFLXECeE5nZYIqBETJLfgQCQiADcag00B0UggAAVEAALZrRQ6JfHSBpAA):
```typescript
type KeysOf<T, Key extends keyof T = keyof T> = Key extends string
    ? T[Key] extends (infer U)[]
      ? `${Key}.${KeysOf<U>}`
      : T[Key] extends Record<string, any>
        ? `${Key}.${KeysOf<T[Key]>}`
        : Key
    : never;
```
(not thoroughly tested!)

Is this in scope for these changes? Or are we not going to allow users to provide nested types and have them as unknowns? (haven't dived into this deeply, so not fully informed sorry!)

bnjmnt4n:
Hey @soedirgo! That repo was actually largely very experimental, and I didn't fully manage to integrate it into a usable Supabase client yet, so this is pretty exciting! On the other hand, while experimenting with that, I did have some notes:

- Using template strings still isn't an ideal DX, because you can't trigger autocompletion of fields. Using objects might allow for better Intellisense autocompletion. Perhaps an ideal solution would be to accept both strings and objects.
- As you mentioned, the cardinality information would allow for the typings to be much more useful. Given that this is a breaking change anyway, it would be cool for the type generation APIs to perhaps provide more information about the tables in the future (possibly separate from the OpenAPI-based solution?) to allow for much better type informations when embedded resources are requested.

src/PostgrestFilterBuilder.ts
soedirgo: The implementation in https://github.com/bnjmnt4n/supabase-client/blob/a1bd9d86fc04ffe9fd5172667644755d96456e29/src/postgrest/PostgrestFilterBuilder.ts#L39 actually does this, but I wanted to keep this simple and fall back to `unknown` for now.

I noticed that it's *not* falling back to `unknown` though - will update the PR. Thanks for noticing!

soedirgo:
> Using template strings still isn't an ideal DX, because you can't trigger autocompletion of fields. Using objects might allow for better Intellisense autocompletion

@bnjmnt4n I agree, after implementing this I concluded that using template strings for the `select` query is still suboptimal. There was actually a suggestion to use arrays instead: https://github.com/supabase/postgrest-js/issues/265. Atm the ideal DX imo is an array of typed objects, each object representing a field.

But I do think that this approach is a good local maximum at least - I'm thinking of not changing the API too much right now (there are other breaking changes in separate PRs alongside this one) - so the plan is to implement the typed array approach in the next major version.

> the cardinality information would allow for the typings to be much more useful. Given that this is a breaking change anyway, it would be cool for the type generation APIs to perhaps provide more information about the tables in the future (possibly separate from the OpenAPI-based solution?) to allow for much better type informations when embedded resources are requested

Yup, I'm currently working on the typegen - which can be run using the Supabase CLI - to complement this PR. (Note that the typing changes here are incompatible with the OpenAPI-based typings.) It'll be based on https://github.com/supabase/postgres-meta, so we can get pretty much any info we want from the database, though I decided to put it off for now since even with FK info determining cardinality is still nontrivial (I probably need to look into how PostgREST does it).

src/PostgrestClient.ts
```diff
@@ -2,12 +2,20 @@ import PostgrestQueryBuilder from './PostgrestQueryBuilder'
 import PostgrestFilterBuilder from './PostgrestFilterBuilder'
 import PostgrestBuilder from './PostgrestBuilder'
 import { DEFAULT_HEADERS } from './constants'
-import { Fetch } from './types'
+import { Fetch, GenericSchema } from './types'
 
-export default class PostgrestClient {
+export default class PostgrestClient<
+  Database = any,
```
J0, line 8: Guessing we'll have to live with these two `any` types because `unknown` would limit the operations we can perform on `Database` ?

J0 (approved):
LGTM

src/PostgrestClient.ts
soedirgo: yeah, it's an escape hatch to ease the transition to the new typings

PR #293, opened 2022-08-03
Adds `explain()` and `geojson()` transforms

steve-chavez:
Adds support for:

- explain: https://github.com/PostgREST/postgrest/pull/2368
- geojson: https://github.com/PostgREST/postgrest/pull/1564

### DIFF
--- a/src/lib/PostgrestTransformBuilder.ts
+++ b/src/lib/PostgrestTransformBuilder.ts
@@ -120,4 +120,12 @@ export default class PostgrestTransformBuilder<T> extends PostgrestBuilder<T> {
     this.headers['Accept'] = 'text/csv'
     return this as PromiseLike<PostgrestSingleResponse<string>>
   }
+
+  /**
+   * Set the response type to GeoJSON.
+   */
+  geojson(): PromiseLike<PostgrestSingleResponse<Record<string, unknown>>> {
+    this.headers['Accept'] = 'application/geo+json'
+    return this as PromiseLike<PostgrestSingleResponse<Record<string, unknown>>>
+  }
 }
--- a/src/lib/PostgrestTransformBuilder.ts
+++ b/src/lib/PostgrestTransformBuilder.ts
@@ -1,4 +1,4 @@
-import { PostgrestBuilder, PostgrestMaybeSingleResponse, PostgrestSingleResponse } from './types'
+import { PostgrestBuilder, PostgrestMaybeSingleResponse, PostgrestResponse, PostgrestSingleResponse } from './types'
 
 /**
  * Post-filters (transforms)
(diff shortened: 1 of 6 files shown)


test/transforms.ts
```diff
@@ -101,3 +101,124 @@ test('abort signal', async () => {
...
+      },
+    ]
+  `)
+})
+
+test('explain with options', async () => {
+  const res = await postgrest
+    .from('users')
+    .select()
+    .explain({ verbose: true, settings: true })
```
steve-chavez, line 175: Here I wanted to test adding `.single()`, like

```js
  const res = await postgrest
    .from('users')
    .select()
    .single()
    .explain()
```

But I got

```
test/transforms.ts:176:6 - error TS2339: Property 'explain' does not exist on type 'PromiseLike<PostgrestSingleResponse<any>>'.
```

This works when running on `node` though.

Any advice on how to make it work? (@soedirgo) 

---

This is necessary because the plan can change depending on the transform.`geojson()` or `csv()` change the generated query.

soedirgo: `.single()` is meant to be called last, not followed by another method - it was the easiest way to implement the typings.

Given both `.single()` and `.geojson()` sets `Accept`, we don't need to support having both at the same time, no?

src/lib/PostgrestTransformBuilder.ts
```diff
@@ -120,4 +120,51 @@ export default class PostgrestTransformBuilder<T> extends PostgrestBuilder<T> {
     this.headers['Accept'] = 'text/csv'
     return this as PromiseLike<PostgrestSingleResponse<string>>
   }
+
+  /**
+   * Set the response type to GeoJSON.
+   */
+  geojson(): PromiseLike<PostgrestSingleResponse<string>> {
```
soedirgo, line 127: Can we have the proper type here? Doesn't look like `data` is `string` looking at the test

src/lib/PostgrestTransformBuilder.ts
```diff
@@ -120,4 +120,51 @@ export default class PostgrestTransformBuilder<T> extends PostgrestBuilder<T> {
...
+    settings = false,
+    buffers = false,
+    wal = false,
+  }: {
+    analyze?: boolean
+    verbose?: boolean
+    settings?: boolean
+    buffers?: boolean
+    wal?: boolean
+  } = {}): PromiseLike<PostgrestSingleResponse<string>> {
```
soedirgo, line 153: also here

soedirgo (approved):

test/transforms.ts
soedirgo: saw the comment in the `.explain()` method - I think it's acceptable to not allow doing both for now, we can think about how to support both after launch week

soedirgo:
Can you give some background on what `.geojson()` does? `.explain()` is simple enough, but how does `.geojson()` change the response? Does it change only geo columns, leaving all other columns the same? Does it only return geo columns?

Also, does `.explain()` always return stuff in the same format? Looks like `Settings` is missing if `options=` isn't specified?

steve-chavez:
> Can you give some background on what .geojson() does? 
> but how does .geojson() change the response?

It runs `st_asgeojson`(postgis func) over the whole row making it follow the structure on https://geojson.org/. 

> Does it change only geo columns, leaving all other columns the same? Does it only return geo columns?

Changes the geocolumns and all the other columns are stored into the `properties` key as per the spec. If more than one geometry column exists on the relation, `st_asgeojson` will only put the first one on the `geometry` key and the other ones will be put on the `properties` key(can be later enhaced to put all into a multipolygon geometry). This behavior can be changed with `select` as one would expect.

> Also, does .explain() always return stuff in the same format? Looks like Settings is missing if options= isn't specified?

Depends on the options specified, more options add more json keys. These follow the same format as doing an [EXPLAIN (FORMAT JSON)](https://www.postgresql.org/docs/current/sql-explain.html). (Note: the format varies across pg versions)

soedirgo:
Gotcha, I guess for now the return type can be:
- `.explain()`: `PromiseLike<PostgrestResponse<Record<string, unknown>>>`
- `.geojson()`: `PromiseLike<PostgrestSingleResponse<Record<string, unknown>>>`

test/transforms.ts
steve-chavez: > .single() is meant to be called last, not followed by another method - it was the easiest way to implement the typings.

I see, I was trying to make `explain()` the last method to be called, it can be done after a `.csv()` or `.single()`.

> saw the comment in the .explain() method - I think it's acceptable to not allow doing both for now, we can think about how to support both after launch week

Ok, agree.

steve-chavez:
@soedirgo Done. Could we make a new release to add these methods to supabase-js v2?

soedirgo:
Will do! v2 isn't released yet (it's on the `next` branch) but I'll add the features there as well

soedirgo (approved):

dshukertjr:
Wow, this seems like a very awesome feature!

PR #296, opened 2022-08-15
feat: add rollback parameter

steve-chavez:
Rollback was added back in https://github.com/supabase/postgrest-js/pull/244 but it's a global parameter. This PR adds a rollback that can be enabled per request - for mutations and rpc(`select` is possible but not that useful).

rollback is needed for EXPLAIN ANALYZE(since it executes the query):

```js
const res = await postgrest
    .from('users')
    .delete({rollback: true})
    .explain({analyze: true})
    .then((res) => res.data)
```


steve-chavez:
Wrong branch, prefer https://github.com/supabase/postgrest-js/pull/297

PR #297, opened 2022-08-15
feat: add rollback parameter to mutations and rpc

steve-chavez:
Rollback was added back in https://github.com/supabase/postgrest-js/pull/244 but it's a global parameter. This PR adds a rollback that can be enabled per request - for mutations and rpc(`select` is possible but not useful, so it's not included).

rollback is needed for EXPLAIN ANALYZE(since it executes the query):

```js
const res = await postgrest
    .from('users')
    .delete({rollback: true})
    .explain({analyze: true})
```

### DIFF
--- a/src/PostgrestClient.ts
+++ b/src/PostgrestClient.ts
@@ -71,6 +71,7 @@ export default class PostgrestClient<
    * @param options  Named parameters.
    * @param options.head  When set to true, no data will be returned.
    * @param options.count  Count algorithm to use to count rows in a table.
+   * @param rollback  Rollback the operation
    */
   rpc<
     FunctionName extends string & keyof Schema['Functions'],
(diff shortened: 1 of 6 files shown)


soedirgo:
@steve-chavez is this only for `EXPLAIN ANALYZE`? Can we automatically rollback just for that query, instead of adding the `rollback` param?

soedirgo:
i.e. add the `rollback` param only in `.explain()` instead

steve-chavez:
> i.e. add the rollback param only in .explain() instead

Hm, rollback on mutations might be useful for test environments as well. 

> i.e. add the rollback param only in .explain() instead

Overall I'm not sure, what would be the drawback of having it outside of explain?

soedirgo:
IMO having it as a parameter for mutations makes the methods feel very cluttered. Alternatively we could have it as a `.rollback()` method instead in `PostgrestTransformBuilder`, the same way we removed `returning: 'minimal'` etc. in favor of `.select()`.

steve-chavez:
@soedirgo Yeah, a `rollback()` method sounds good to me. In fact I did that first but I thought it didn't fit the "transform" module.

soedirgo:
:+1: will do it in a separate PR.

PR #301, opened 2022-08-17
feat: add text format to explain()

steve-chavez:
json format for explain was added in https://github.com/supabase/postgrest-js/pull/293 but it's hard to read. This PR adds a text format so it looks like:

```js
postgrest.from('users').select().explain({format: 'text'}).then(({ data, error }) => console.log(data, error))

Aggregate  (cost=17.65..17.68 rows=1 width=112)
  ->  Seq Scan on users  (cost=0.00..15.10 rows=510 width=132)
```

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -156,6 +156,7 @@ export default class PostgrestTransformBuilder<
    * @param settings  If `true`, include information on configuration parameters that affect query planning.
    * @param buffers  If `true`, include information on buffer usage.
    * @param wal     If `true`, include information on WAL record generation
+   * @param format  The format of the output, can be 'json'(default) or 'text'
    */
   explain({
     analyze = false,
@@ -163,13 +164,17 @@ export default class PostgrestTransformBuilder<
     settings = false,
     buffers = false,
     wal = false,
+    format = 'json',
   }: {
     analyze?: boolean
     verbose?: boolean
     settings?: boolean
     buffers?: boolean
     wal?: boolean
-  } = {}): PromiseLike<PostgrestResponse<Record<string, unknown>>> {
+    format?: 'json' | 'text'
+  } = {}):
+    | PromiseLike<PostgrestResponse<Record<string, unknown>>>
+    | PromiseLike<PostgrestSingleResponse<string>> {
     const options = [
       analyze ? 'analyze' : null,
       verbose ? 'verbose' : null,
@@ -183,7 +188,8 @@ export default class PostgrestTransformBuilder<
     const forMediatype = this.headers['Accept']
     this.headers[
       'Accept'
-    ] = `application/vnd.pgrst.plan+json; for="${forMediatype}"; options=${options};`
-    return this as PromiseLike<PostgrestResponse<Record<string, unknown>>>
+    ] = `application/vnd.pgrst.plan+${format}; for="${forMediatype}"; options=${options};`
+    if (format == 'json') return this as PromiseLike<PostgrestResponse<Record<string, unknown>>>
+    else return this as PromiseLike<PostgrestSingleResponse<string>>
   }
 }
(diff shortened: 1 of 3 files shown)


darora (approved):

src/PostgrestTransformBuilder.ts
```diff
@@ -183,7 +188,8 @@ export default class PostgrestTransformBuilder<
     const forMediatype = this.headers['Accept']
     this.headers[
       'Accept'
-    ] = `application/vnd.pgrst.plan+json; for="${forMediatype}"; options=${options};`
-    return this as PromiseLike<PostgrestResponse<Record<string, unknown>>>
+    ] = `application/vnd.pgrst.plan+${format}; for="${forMediatype}"; options=${options};`
+    if (format == 'json') return this as PromiseLike<PostgrestResponse<Record<string, unknown>>>
```
darora, line 192:
```suggestion
    if (format === 'json') return this as PromiseLike<PostgrestResponse<Record<string, unknown>>>
```

kiwicopple:
I'm wondering how the developer workflow for this will be?

If I was building an app, I would probably want to do something like:

```js
createClient('URL', 'KEY', {
  db: {
    withQueryAnalyzer: process.env.ENVIRONMENT == 'test' ? 'explain' : null // leave room for "explain analyze"?
  }
})

const { data, error, plan } = await supabase.from('table').select()

console.log('plan', plan)
```

I think the key is that developers won't want to edit their code _a lot_ to get both the plan and the data, but it looks like PostgREST returns either an explain or a request? (which would make this workflow difficult)

Another workflow that they could adopt is


```js

const query = supabase.from('table').select()

const { data, error } = await query

if (process.env.ENVIRONMENT != 'production') {
  const { data } = await query.explain()
  console.log('query plan', data)
}

```

Not sure what the best practice would be here. This is a really awesome feature, and I think it will be very powerful if we can build a simple workflow around it

steve-chavez:
Main idea is to allow users to check how their RLS policies are affecting JS queries, from the JS client itself. Otherwise this requires some extra steps on the SQL editor; doable as shown on the [slip blog post](https://www.slip.so/tutorials/row-level-security-performance-in-supabase) but doing it from JS provides quicker feedback.

> I'm wondering how the developer workflow for this will be?

I thought basically:

```js
const { user, session, error } = await supabase.auth.signIn({
  email: 'example@email.com',
  password: 'example-password',
})

const { data, error } = await supabase.from('table').select().explain({analyze: true})

console.log(data)
// check the plan, if the query takes too long
```

> 'explain' : null // leave room for "explain analyze"?

Note that this is already possible with `explain({analyze: true})`. Can also be paired with [rollback](https://github.com/supabase/postgrest-js/pull/297).

> const { data, error, plan } = await supabase.from('table').select

Using a different key for the plan seems like a good idea, like what we do for `count`.

> but it looks like PostgREST returns either an explain or a request? (which would make this workflow difficult)

Some issues with that interface:

- PostgREST would need to execute 2 queries for the request and that would break the promise of "1 request = 1 query" thus notably reducing performance.
- Doing EXPLAIN ANALYZE would mean executing the same query twice which would be surprising behavior.
- The EXPLAIN output would be too large(in JSON format specially) to send in a response header.

So for this feature I believe it's better to stick to PostgreSQL behavior: a query cannot get data when using EXPLAIN.

> Not sure what the best practice would be here. This is a really awesome feature, and I think it will be very powerful if we can build a simple workflow around it

Not sure either. I think it as the first step towards better observability for executions plans, so definitely not final.

steve-chavez:
> I'm wondering how the developer workflow for this will be?

For me, the ideal workflow would involve using `supabase-js` in the browser console. There I'd be able to pretty print the data with [console.table](https://mariusschulz.com/blog/advanced-javascript-debugging-with-console-table), copy the execution plan with just the `copy(plan)` function.

I've been trying to do that but seems I need browserify and additional stuff for getting a minified js library.

It would be great if `supabase-js` could be included when visiting `app.supabase.com/project/_/api`.

steve-chavez:
> const { data, error, plan } = await supabase.from('table').select
> Using a different key for the plan seems like a good idea, like what we do for count.

I'll move the plan from `data` to `plan` in another PR.

Issue #295, opened 2022-08-08
`error` is returned as `[]` if `.update()` matched no records


p-schneid:
# Feature request

## Is your feature request related to a problem? Please describe.

This is not a feature, so much as an improvement. I noticed when preforming an update, the response returns an error if the request didn't match any records. This is expected. However the response.error comes back as a empty array, which is not very helpful. A message could make error handling and debugging easier for consumers. 

## Describe the solution you'd like

A message indicating the request did not match any existing records

## Describe alternatives you've considered

I stop whining about missing error messages. After all, the response does return a 404. 

## Additional context

![image](https://user-images.githubusercontent.com/33016331/183513652-7401e5a5-8e3e-4445-926c-fa0e2691a848.png)

soedirgo:
Thanks! This seems to be a bug. Can you provide the supabase-js version you used?

steve-chavez:
Related https://github.com/PostgREST/postgrest/issues/2343

steve-chavez:
Just to be clear, on this issue there is no error message in the response body coming from the server, only a 4xx status. This needs a fix on the PostgREST side.

PR #379, opened 2022-12-16
fix: wrong response on update w/o match

soedirgo:
Fixes https://github.com/supabase/postgrest-js/issues/295

### DIFF
--- a/src/PostgrestBuilder.ts
+++ b/src/PostgrestBuilder.ts
@@ -106,9 +106,23 @@ export default abstract class PostgrestBuilder<Result>
 
         try {
           error = JSON.parse(body)
+
+          // Workaround for https://github.com/supabase/postgrest-js/issues/295
+          if (Array.isArray(error) && res.status === 404) {
+            data = []
+            error = null
+            status = 200
+            statusText = 'OK'
+          }
         } catch {
-          error = {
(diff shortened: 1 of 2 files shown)


src/PostgrestBuilder.ts
```diff
@@ -106,9 +106,23 @@ export default abstract class PostgrestBuilder<Result>
 
         try {
           error = JSON.parse(body)
+
+          // Workaround for https://github.com/supabase/postgrest-js/issues/295
+          if (Array.isArray(error) && res.status === 404) {
```
dshukertjr, line 111: @soedirgo 
I was working on adding this update to the Flutter library and have some questions. 

What is this `Array.isArray(error)` check for? Is checking `res.status === 404` not enough in this case?

steve-chavez: I think that patch is no longer necessary even on `postgrest-js` since https://github.com/PostgREST/postgrest/issues/2343 is fixed and already deployed to all projects.

dshukertjr: @steve-chavez 
That is amazing! Was this deployed very recently? I had a user report this 2 days ago [here](https://github.com/supabase/supabase-flutter/issues/315).

PR #380, opened 2022-12-17
Extend query parser w/basic support for `->`/`->>` JSON operators

mqp:
Postgres supports [accessing properties in JSON and JSONB columns](https://www.postgresql.org/docs/current/functions-json.html) by using the `->` and `->>` operators. Postgrest [supports this too](https://postgrest.org/en/stable/api.html#json-columns). However, trying to use this in the Supabase `select` API makes Typescript barf:

```typescript
/* schema: users: { Row: { data: Json } } } */
const { data } = await db.from('users').select('data->username, data->>numLikes')

// inferred type of `data` is `ParserError<'Unexpected input: ->username, data->>numLikes'>[] | null`
```

This small extension to the query parser types makes simple `column->foo->bar->>baz` property chains work properly, e.g. the above query will infer a type of `{ username: Json, numLikes: string }`. 

Note that this doesn't currently work in combination with embedded fields or renaming -- the query parser types would realistically need to be refactored to accomplish that without a serious amount of cut-and-pasted code. I aimed to just hit the common case and scratch my own itch with minimal effort here.

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -38,13 +38,7 @@ type Digit = '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '0'
 
 type Letter = Alphabet | Digit | '_'
 
-type Json =
-  | string
-  | number
-  | boolean
-  | null
-  | { [key: string]: Json }
-  | Json[]
+type Json = string | number | boolean | null | { [key: string]: Json } | Json[]
 
 // /**
 //  * Parsed node types.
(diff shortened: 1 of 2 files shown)


mqp:
Do you have any interest in this kind of work? I have a project that makes heavy use of JSON columns, and I need to decide whether to try to contribute JSON column QoL improvements upstream like this, or whether I should maintain a fork.

soedirgo:
Hey, sorry for the late response! Yes, we definitely need this - will take a look later today.

soedirgo (approved):

mqp:
> Hey, sorry for the late response! Yes, we definitely need this - will take a look later today.

<3

PR #390, opened 2023-01-17
Extend typing to make JSON expressions work with `order`

mqp:
Previously, if you wrote a query like this:

```typescript
// presumes a table foos that has a JSONB `data` column
const { data, error } = await client.from('foos').select('id').order('data->>count', { ascending: true })
```

It would fail to typecheck, because the typing for `order` thought that if the first parameter isn't a column name from the `foos` table in the schema, then it must be a foreign table reference, so the `foreignTable` property in the options must be specified.

Now, Typescript can extract the column name from the `data->>count` expression, and check that against the schema.

I think there are other expressions you can use here (the Postgrest [docs](https://postgrest.org/en/stable/api.html#ordering) mention computed columns) but this covers the JSON accessor case. (We could also probably use similar code in the filter builder to get better typechecking when using JSON accessors in a variety of methods.)

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -7,6 +7,10 @@ import {
   PostgrestSingleResponse,
 } from './types'
 
+/* Extract the column name from a potential JSON accessor expression. */
+type ColumnName<T extends string> =
+  T extends `${infer TName}->${infer _TRest}` ? TName : T
+
 export default class PostgrestTransformBuilder<
   Schema extends GenericSchema,
   Row extends Record<string, unknown>,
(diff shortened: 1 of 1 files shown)


mandarini:
Hi @mqp ! Thanks for the contribution and your patience.

This repository is deprecated and has moved to the new [Supabase JS monorepo](ttps://github.com/supabase/supabase-js). Since this PR has been inactive for over a year, I’m going to close it to keep the old repo tidy, before archiving.

If you believe this change is still needed, please open a new PR in the monorepo and include a link back to this thread for context:

Note: This old repository will be archived on October 10, 2025. Thank you again for your effort and understanding!

PR #394, opened 2023-02-04
fix: rpc return type

### LINKED ISSUE #387: rcp() return type does not match SQL function return type
# Bug report

I have a SQL function that returns a boolean  and Supabase CLI generates the TypeScript types as expected:
```ts
Functions: {
    some_function: {
        Args: Record<PropertyKey, never>;
        Returns: boolean;
    };
}
```

But when I call the function, `data` has the type `boolean[] | null`, while I would expect `boolean | null`.
```ts
const { data } = supabaseClient.rpc('some_function')
```

soedirgo:

Fixes #387.

RPC return types aren't getting propagated properly. E.g. if a function returns `boolean`, the `data` will be typed as `boolean[]`.

To fix this, we change PostgrestBuilder's type parameter to be the actual type of `data`, instead of the element type of `data` (which assumes `data` is always an array, which is incorrect).

In doing so we effectively remove all usage of `PostgrestResponse` and `PostgrestMaybeSingleResponse`. We can stop exporting these on the next major version.

This also fixes the behavior of `.returns<NewResult>()`, which was inconsistent with the documentation (i.e. `data` was typed as `NewResult[]` instead of `NewResult`).


steve-chavez (approved):

J0 (approved):

natedunn:
oof, i was wondering why all my types got borked overnight.

soedirgo:
Sorry about that @natedunn - do you mind sharing how it's breaking your types? While we consider it a bug here we do want to make sure the fix isn't causing more issues at least.

natedunn:
@soedirgo  Yeah I can understand the mentality there, so no worries. 

There's not much to describe in terms of my issues. The PR description laid it out pretty well, so I knew what I had to change: 

This stopped working
```
  .returns<SomeDefinedType>()
```

And this fixed the issue
```
  .returns<SomeDefinedType[]>()
```

bombillazo:
I created an issue in the other repo: https://github.com/supabase/postgrest-js/issues/430

Probably related to this.

PR #396, opened 2023-02-07
Export PostgrestResponseSuccess and PostgrestResponseFailure

kjrocker:

Type change. Discussion here: https://github.com/supabase/supabase/discussions/12228

Only the top-level `PostgrestResponse` types are exported.

## What is the new behavior?

Export the sub-types that go into `PostgrestResponse` for type-union/guard-clause purposes.

## Additional context

I noticed the plan is to rename `PostgrestSingleResponse` and deprecate the current `PostgrestResponse`, so instead of exporting `PostgrestSingleResponseSuccess` under its current name I went with `PostgrestResponseSuccess`.

This _might_ mean holding the change back until v3, but since its to a (formely) internal type we should be okay.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,6 +5,8 @@ export { default as PostgrestTransformBuilder } from './PostgrestTransformBuilde
 export { default as PostgrestBuilder } from './PostgrestBuilder'
 export {
   PostgrestResponse,
+  PostgrestResponseFailure,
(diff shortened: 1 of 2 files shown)


mqp:
This would be convenient for me too.

soedirgo (approved):
Sgtm, thanks!

PR #397, opened 2023-02-11
Explain renaming of returned columns

yannxaver:
I could not find a note about it in the docs so far


steve-chavez (approved):
Thanks for the contrib!

Issue #361, opened 2022-10-31
Error logged when using `.maybeSingle()` returns no result


probablykasper:
# Bug report

## Describe the bug

When selecting using `.maybeSingle()`, I'm getting an error logged to console if there's 0 rows returned. There's no error if I remove `.maybeSingle()`, and there's no error if a row is returned.

To be clear, the query still returns successfully.

![image](https://user-images.githubusercontent.com/11315492/199113146-911bb145-c6f3-40e3-ae05-854b22dfd1a8.png)

The stack trace points to this line (probably badly source mapped):
https://github.com/supabase/postgrest-js/blob/84f151834a4e305508cd246cba9ea33e6affc0f2/src/PostgrestBuilder.ts#L102

## To Reproduce

```js
	const profileResult = await supabaseClient
		.from('profiles')
		.select()
		.eq('id', session.user.id)
		.maybeSingle()
```

## System information

- OS: macOS
- Browser: Brave
- Version of supabase-js: 2.0.4
- Version of Node.js: 16.15.0

soedirgo:
> The stack trace seems to points to this line (probably a bad source map)

Hmm the trace points to `PostgrestBuilder.ts:70` which looks right:

https://github.com/supabase/postgrest-js/blob/84f151834a4e305508cd246cba9ea33e6affc0f2/src/PostgrestBuilder.ts#L70

So it comes from the `fetch` invocation. I suspect the browser's native `fetch` is logging all `4xx`s - the way we do `maybeSingle()` is we do `single()` but ignore the error, so the `406` comes from PostgREST. Should be harmless, but noisy - I suppose we could instead treat it as a non-`single()` and just throw if there are >1 rows.

steve-chavez:
> the way we do maybeSingle() is we do single() but ignore the error, so the 406 comes from PostgREST

Not ideal to ignore the error on the client as the logs would also still contain the 406. I think https://github.com/PostgREST/postgrest/pull/2164#issuecomment-1165035270 would be the ideal solution.

(basically a header that defines expected number of returned rows)

soedirgo:
Gotcha, changing this issue to track refactoring the code to use that feature once it's out :+1:

brentrobbins:
Hi @soedirgo Any update on when this might be addressed?

altechzilla:
This has been driving me insane to debug on Vercel. Any updates here?

soedirgo:
[...]

PR #424, opened 2023-05-11
fix: partial fix for maybeSingle logging 406s

soedirgo:
Partial fix for https://github.com/supabase/postgrest-js/issues/361


thorwebdev (approved):
LGTM

Issue #370, opened 2022-11-16
Query parser can't parse `id::text`


ozanmakes:
# Bug report

## Describe the bug

I have BigInt id columns that I have to query with `id::text` everywhere due to overflow issues (related: #319). However, these queries breaks the `ParseQuery` type.

## To Reproduce

Backed by a column of type `BigInt`, write a query and typecheck with TypeScript:

```ts
  let { data, error } = await supabase
      .from("subscriptions")
      .select("id");
```

Hover over `data` to see the inferred type.

## Expected behavior

As a result of the query above, `id` type is computed as `string`.

## Screenshots
<img width="741" alt="image" src="https://user-images.githubusercontent.com/111265/202151538-7eda23c6-4e01-4872-9687-90812f1e9aa6.png">
<img width="440" alt="image" src="https://user-images.githubusercontent.com/111265/202151616-3e0f257a-2d58-45c0-8692-13818a9e3b4b.png">

## System information

- OS: macOS
- Version of @supabase/postgrest-js: 1.1.0
- Version of @supabase/supabase-js: 2.1.0
- Version of Node.js: v16.14.0

PR #429, opened 2023-05-25
Add support for type casting (id::text)

ozanmakes:

This change adds support for type casting in select queries. For example an id column with the type bigint can be casted to text by "id::text".

Equivalent PostgreSQL type for TypeScript type is based on the type generator [pgTypeToTsType](https://github.com/supabase/postgres-meta//blob/80809c71d78d875524d7d9865c2458136207aa10/src/server/templates/typescript.ts#L421) function.

Fixes https://github.com/supabase/postgrest-js/issues/370

## What is the new behavior?

- `id::text` will cast a `bigint` column to `string` in TypeScript.
- `_text` will be `text[]`.
- `newname:id::text` will also work

## Additional context

This is by no means a complete list of all possible type casts nor is it sound, but this is much less error prone than having to do an unsafe cast such as 
```ts
supabase.from("users").select("id::text").returns<{ id: string }[]>
```

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -40,6 +40,68 @@ type Letter = Alphabet | Digit | '_'
 
 type Json = string | number | boolean | null | { [key: string]: Json } | Json[]
 
+type SingleValuePostgreSQLTypes =
+  | 'bool'
+  | 'int2'
+  | 'int4'
+  | 'int8'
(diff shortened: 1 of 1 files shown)


osener:
Hi @soedirgo. Any chance this can be reviewed?

steve-chavez (approved):

PR #436, opened 2023-06-10
Type change for columns that do not exist in the schema

3ru:

The type for columns that do not exist in the schema, which used to be `? ` to `null`.

Currently, given a column that does not exist, the property can still be specified.

image

## What is the new behavior?
If a non-existent column is given, the user cannot specify a non-existent column in order to return null.

image

This makes it easier to recognize column name problems from why they are null or from errors.

## Additional context

It may be that this should return an error type similar to `ParserError` , `GenericStringError`, etc., but I think that would greatly expand the area to be modified.

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -136,7 +136,9 @@ type ConstructFieldDefinition<
         : never
     }
   : Field extends { name: string; original: string }
-  ? { [K in Field['name']]: Row[Field['original']] }
+  ? Field['original'] extends keyof Row
+    ? { [K in Field['name']]: Row[Field['original']] }
+    : never
   : Field extends { name: string; type: infer T }
   ? { [K in Field['name']]: T }
   : Record<string, unknown>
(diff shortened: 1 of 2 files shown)


3ru:
@soedirgo 
Beautiful🙌
Thank you and merging is fine!

soedirgo (approved):

lopezjurip:
It broke custom calculated columns using psql functions

Here is an example:

```sql
create table profiles (
  id uuid not null primary key default uuid_generate_v4(),
  user_id uuid unique references auth.users (id) on delete cascade not null,
);

create or replace function _email(this profiles)
  returns text strict stable
  language plpgsql
  as $$
    begin
      return (select users.email from auth.users as users where users.id = this.user_id and users.id = auth.uid() limit 1);
    end;
  $$;
```

```typescript
const { data } = await supabase.from("profiles").select("id, email: _email");
// data type is: SelectQueryError<"Referencing missing column `_email`">
```

Previous version just declared `email` as unknown and I could overwrite that later.

PR #439, opened 2023-06-22
Documentation improvements for the `range()` method

pangolingo:

Docs update

These updates are based on feedback in the Supabase repo's issue #15222 (https://github.com/supabase/supabase/issues/15222)

The original docs state that "range" takes a starting index and ending index and returns the values between them. This implies that it's using the primary ID. But it actually works more like limit + offset.

## What is the new behavior?

I've updated the docs to explain how it works in greater detail.

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -101,7 +101,7 @@ export default class PostgrestTransformBuilder<
   }
 
   /**
-   * Limit the query result by starting at an offset (`from`) and ending at the offset `from + to`).
+   * Limit the query result by starting at an offset (`from`) and ending at the offset (`from + to`).
    * Only records within this range are returned.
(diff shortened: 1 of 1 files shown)


src/PostgrestTransformBuilder.ts
```diff
@@ -101,7 +101,11 @@ export default class PostgrestTransformBuilder<
   }
 
   /**
-   * Limit the query result by `from` and `to` inclusively.
+   * Limit the query result by starting at an offset (`from`) and ending at the offset `from + to`).
```
soedirgo, line 104:
```suggestion
   * Limit the query result by starting at an offset (`from`) and ending at the offset (`from + to`).
```

soedirgo:
Thanks for the PR!

soedirgo (approved):

Issue #280, opened 2022-06-10
Method to set schema on a call by call basis.


GaryAustin1:
# Feature request

Right now you can only set a schema in initial supabase client setup.
It would be very useful to be able to override the default ('public' normally) schema on individual calls.

## Describe the solution you'd like
Add a postgrest-js operation  .schema('myschema') that can be used with other operations on a single call basis.

## Describe alternatives you've considered
Currently the below works in my environment.  It does not impact storage and auth, but have not tested on realtime yet.
```
let schema = ''
var supabase = supabase.createClient(SUPABASE_URL, SUPABASE_KEY, {
    fetch: (requestInfo, requestInit) => {
        if (schema) requestInit.headers['Accept-Profile'] = schema
        return(fetch(requestInfo,requestInit))
    }
```
Then you do the following for a different schema
```
    schema = 'storage'
    const response = await supabase
        .from('objects')
        .select('*')
    schema = ''
```
For modular code one could add 
```
export function setSchema(schemaName) {
 schema = schemaName
}
```
And call that before and after supabase database call to another schema.
## Additional context

It appears the only thing required to make PostgREST work with an API available schema is setting the Accept-Profile' header.  This should be easily done in postgrest-js with a .schema('myschema') function added to postgrest calls.
```
    const response = await supabase
        .from('objects')
        .schema('storage')
        .select('*')
```

GaryAustin1:
A could swear I tested the code below before, and it did not pass down to postgrest-js after the initial setup, but seems to work (at least for javascript, not sure what typescript would complain about).  
```
supabase.schema = 'storage'
await supabase.from('objects').select('*')
supabase.schema = 'public' // put it back
```
I don't find this documented anywhere, but it seems to work now...

Anyway to get the official way if any to toggle schema for different requests?

soedirgo:
+1 on using the builder pattern to switch schemas, though I'd put it before the `.from()`, e.g.

```ts
    const response = await supabase
        .schema('storage')
        .from('objects')
        .select()
```

steve-chavez:
Might be worth thinking how the interface would look like when postgREST supports embedding on different schemas(just needs an additional header semantic). Maybe like:

```js
const { data, error } = await supabase
  .schema({ countries: "core", cities: "protected" })
  .from('countries')
  .select(`
    name,
    cities (
      name
    )
  `)
```

I guess extending the `.schema()` method to take an object later on shouldn't be a problem.

soedirgo:
[...]

PR #441, opened 2023-06-26
Optionally set schema per PostgrestQueryBuilder

bdotsamir:

This introduces a new method to the PostgrestQueryBuilder file that allows the developer to specify which schema they're targeting with the table specified in `.from()`.

Currently, there is no way to do this. If I want to have only one instance of supabase-js, I can only set its schema once- upon creation. If, somewhere else in the code, I'd like to perform an operation on a table that's in a schema *other* than the one I specified, I'm SOL (or: I have to create a new supabase-js instance on the fly and set *its* schema, then run the query, which is ugly).

Related issues:
* https://github.com/supabase/postgrest-js/issues/280

## What is the new behavior?

```tsx
    supabase.from("test_table")
      .setSchema("test_schema") // ✨ :D
      // ^ I can now set the schema for this query.
      .select("*")
      .then(({ data, error }) => {
        console.log('test_schema.test_table data', data);
        console.log('test_schema.test_table error', error);
      });

    supabase.from("test_public_table")
      .select("*")
      // Note that it is optional, and does not affect other queries.
      .then(({ data, error }) => {
        console.log('public.test_public_table data', data);
        console.log('public.test_public_table error', error);
      });
      
```

Sample output:
image

## Additional context

Clearly, it's not just this class that needs to change. It's also the typings, and with it possibly updating the [database definition generator](https://supabase.com/docs/reference/javascript/typescript-support) to include *all* public schemas / all schemas the anon key has access to (via the project API settings, see image below). This then requires changing the way clients _accept_ database typings (or, actually *any* place a `Database` typings file is accepted, e.g. `createPages{Server,Browser}Client<Database>()`, etc.). More to come later.

Also, when I created this new schema for testing purposes, [I had to grant it access to the roles I wanted](https://supabase.com/docs/guides/integrations/prisma#missing-grants). It would be nice if this were an automated thing? Possibly even further scope creep, I'm not gonna push it too much.

image

By the way, the way that I tested this was not directly through `new PostgrestClient(URL)`, I basically `npm link`'d the HELL out of everything. My development cycle looked like:
1. Clone postgrest-js and supabase-js
2. Create a test project (next.js in this case)
3. Make the changes to `postgrest-js`, then `npm link` it so I can access this specific version of the package anywhere
4. In supabase-js, run `npm link @supabase/postgrest-js` to *override* its dependency, instead pointing to my local copy
5. In supabase-js, run `npm link` so I can access *this specific overriden* version anywhere
6. In my testing project, run `npm link @supabase/supabase-js`
7. Repeat steps 3 - 7.

### DIFF
--- a/src/PostgrestQueryBuilder.ts
+++ b/src/PostgrestQueryBuilder.ts
@@ -32,6 +32,11 @@ export default class PostgrestQueryBuilder<
     this.fetch = fetch
   }
 
+  setSchema(schema?: string): PostgrestQueryBuilder<Schema, Relation, Relationships> {
+    this.schema = schema
+    return this
+  }
+
   /**
    * Perform a SELECT query on the table or view.
    *


bdotsamir:
Additionally, if anybody would like to brainstorm this further (off this thread), I have a running [forum post on the supabase discord](https://discord.com/channels/839993398554656828/1122651341999841320). Feel free to ping me there, I'm `Strange#0009`.

bdotsamir:
The grammar of this could be changed too, of course. It could look like:
* `supabase.setSchema('other_schema').from('table')`
* `supabase.from('other_schema', 'table')` or vice versa
   * Maybe even overloading this function so if the user passes in only the table it selects it from `public` by default?

bdotsamir:
I am actually looking for feedback on this PR. It's in a deliberately "unfinished" state because I want to be sure it's been vetted by the project's members.

src/PostgrestQueryBuilder.ts
```diff
@@ -32,6 +32,11 @@ export default class PostgrestQueryBuilder<
     this.fetch = fetch
   }
 
+  setSchema(schema?: string): PostgrestQueryBuilder<Schema, Relation, Relationships> {
```
steve-chavez, line 35: How about a method like the one I proposed before here https://github.com/supabase/postgrest-js/issues/280#issuecomment-1158087393

```js
  .schema({ countries: "core", cities: "protected" })
  .from('countries')
  .select(`
    name,
    cities (
      name
    )
  `)
```

Just `schema` and it should take multiple objects. Otherwise this method will be subject to breaking changes.

steve-chavez: We can have a typescript type that's `string or object` right? If so, maybe we can make the string a global schema for starters.

@soedirgo WDYT?

bdotsamir: What is that syntax you mentioned when calling .schema()? I’m not familiar with the core or protected values

steve-chavez: Those are other schemas.

bdotsamir: Are they aliases? Is “countries” an alias for “core” and “cities” for “protected”?

dshukertjr: @bdotsamir `countries` and `cities` are the table names and `core` and `protected` are the schema names for those tables. This way, the client library can perform joins across different schemas.

bdotsamir: Ah, now I see. Thank you for clarifying. I’ll have a look soon

sjones6:
Proposing an alternative implementation to solve this, but at the supabase-js level: https://github.com/supabase/supabase-js/pull/828

Does not implement the suggestion to use  something like `.schema({ countries: "core", cities: "protected" })` to enable multi-schema joins in a single query.

steve-chavez:
@sjones6 I suggest adding the `schema()` method to `postgrest-js` directly https://github.com/supabase/postgrest-js/issues/280#issuecomment-1666382213.

sjones6:
> @sjones6 I suggest adding the `schema()` method to `postgrest-js` directly [#280 (comment)](https://github.com/supabase/postgrest-js/issues/280#issuecomment-1666382213).

To clarify, which class are you suggesting the addition to: `PostgrestClient` or `PostgrestQueryBuilder`?

Having a `schema(schema: SchemaName)` method on `PostgrestClient` doesn't seem to make as much sense since constructor accepts the schema and changing it for the whole `PostgrestClient` could have unexpected side-effects for other queries.

Without additionally adding a method to the supabase client to proxy to the postgrest-js class, it would have to be:

```ts
await supabase.from('objects').schema('storage').select();
```

and never 

```ts
await supabase.schema('storage').from('objects')
```

It seems somewhat backwards to have the table first and then the schema, but not sure if that's a consideration at all?

soedirgo:
I think it makes sense to add it to `PostgrestClient`, because otherwise you're not able to use `.rpc()` on other schemas.

> changing it for the whole PostgrestClient could have unexpected side-effects for other queries

Agreed, `.schema()` call should leave the object untouched, and instead create a new object with the schema switched (i.e. basically your [PR](https://github.com/supabase/supabase-js/pull/828)).

sjones6:
@soedirgo , raised a new MR to implement the `schema` method on `PostgrestClient` per direction here: https://github.com/supabase/postgrest-js/pull/455

I think then the other [PR](https://github.com/supabase/supabase-js/pull/828) could be modified like so:

```ts
export default class SupabaseClient {
  /* snip */

  /**
   * Perform a query on a schema distinct from the default schema supplied via
   * the `options.db.schema` constructor parameter.
   *
   * The schema needs to be on the list of exposed schemas inside Supabase.
   *
   * @param schema - The name of the schema to query
   */
  schema(
    schema: string & keyof Database
  ): PostgrestClient<Database[typeof schema] extends GenericSchema ? Database[typeof schema] : any, any> {
    return this.rest.schema(schema);
  }
}
```

soedirgo:
Have merged the other PR - thanks for this @bdotsamir!

bdotsamir:
Whoops, sorry it took me a while to get to this. Thanks for letting me know 🫡

PR #443, opened 2023-07-13
Mark array filter values readonly

wyozi:

Mark arrays accepted by filter builder methods readonly.

Currently filter builder methods accept non-readonly arrays. This is fine, but forces casting readonly arrays to non-readonly versions even though Supabase methods do not mutate the passed-in arrays.

See:

```ts
const ALLOWED_STATUSES = ["ACCEPTED", "PENDING"] as const;

supabase.from("tasks")
  .select()
  .eq("status", ALLOWED_STATUSES) // tsc error: Argument of type 'readonly (..)[]' is not assignable to parameter of type '(..)[]'.
```

## What is the new behavior?

All filter builders accepting arrays will accept the readonly version. Since this only changes function parameter signature and none of the returned values, it should be a non-breaking change.

## Additional context

Quite many libraries (and notably, Typescript's own `as const`) use readonly arrays, so this makes working with them a bit easier and less-casting-needed.

### DIFF
--- a/src/PostgrestFilterBuilder.ts
+++ b/src/PostgrestFilterBuilder.ts
@@ -124,28 +124,34 @@ export default class PostgrestFilterBuilder<
     return this
   }
 
-  likeAllOf<ColumnName extends string & keyof Row>(column: ColumnName, patterns: string[]): this
-  likeAllOf(column: string, patterns: string[]): this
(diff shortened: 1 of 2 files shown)


soedirgo (approved):
Thanks!

PR #446, opened 2023-07-20
draft: Support spread in select query

wyozi:

Preliminary support for [spreading resource](https://postgrest.org/en/stable/references/api/resource_embedding.html#spread-embedded-resource).

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -100,6 +100,15 @@ type ConstructFieldDefinition<
   Field
 > = Field extends { star: true }
   ? Row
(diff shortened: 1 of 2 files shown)


src/select-query-parser.ts
```diff
@@ -100,6 +100,15 @@ type ConstructFieldDefinition<
   Field
 > = Field extends { star: true }
   ? Row
+  : Field extends { spread: true; original: string; children: Array<{ name: string }> }
+  ? {
+      [K in Field['children'][number] as K['name']]: ConstructFieldDefinition<
```
wyozi, line 105: Not sure about this. Should probably call `GetResultHelper` instead?

PR #364, opened 2022-11-08
Improve return types for `throwOnError`

Cryrivers:
## What kind of change does this PR introduce?

This PR improves return types for `throwOnError`, which excludes `null` for failure responses in the return type. as `throwOnError` raises exceptions instead of providing a failure response.

## What is the current behavior?

Currently `throwOnError` still types the return data as `Data | null`, whereas the `null` (failure response) case is impossible to happen.

Screenshot 2022-11-08 at 12 34 35 PM

CleanShot 2022-11-08 at 12 28 45@2x

## What is the new behavior?

`throwOnError` just types the return data as `Data` (`null` is still kept if the data is nullable), if error happens, it throws an exception.

Screenshot 2022-11-08 at 12 37 19 PM

Screenshot 2022-11-08 at 12 37 55 PM

## Additional context

This PR addresses https://github.com/supabase/supabase-js/issues/554

### TypeScript issue

TypeScript (4.8, 4.9) complains the circular references, however the types can be correctly inferred in Visual Studio Code.

```
src/PostgrestBuilder.ts:59:19 - error TS2577: Return type annotation circularly references itself.

59   throwOnError(): EnableThrowOnError<this> {
```

PR #454, opened 2023-07-30
fix: throwOnError function call should return response success type always

dogukanakkaya:

This PR improves the typing of `.throwOnError` function. Which should return `PostgrestResponseSuccess<T>` type.

Right now `.throwOnError` is returning nullable data.

## What is the new behavior?

If `.throwOnError` called on the query it will always return `PostgrestResponseSuccess<T>` instead of `PostgrestResponseSuccess<T> | PostgrestResponseFailure`

## Additional context

This is a breaking change because previously you could call other functions after `.throwOnError` because it does not execute the query but only sets the `shouldThrowOnError` property to `true` to be used later on when executing the query.

With this change, chaining more functions after `.throwOnError` is not possible since it executes the query and check the result and throws in case of error or return the `PostgrestResponseSuccess` in case of success.

I saw [this PR](https://github.com/supabase/postgrest-js/pull/364) right after I finished working on that. Which resolves the same issue with Typescript only instead, which I also tried at first since it gives developers more flexibility since we don't necessarily have to execute the query after `.throwOnError`. But looks like that PR is freezed/stuck since Nov 2022 because of a Typescript issue.

Also related with: https://github.com/supabase/supabase-js/issues/801


wyozi:
In my codebase I use a helper called `dataOrThrow`. It does what it says, i.e. do `throwOnError` but also change the return value to the `.data` property. It would be a potential option if avoiding breaking changes was a goal, although I think it would be a good addition to the codebase in the first place

PR #455, opened 2023-08-05
feat: add dynamic schema to postgrest client

sjones6:
Resolves https://github.com/supabase/postgrest-js/issues/280

New feature

See discussion on https://github.com/supabase/postgrest-js/issues/280

The current behavior is that the `PostgrestClient` is created with a schema passed into the constructor. The class instance cannot reference tables or rpc calls for any other schemas unless a new instance is created.

```ts
// implicitly "public" schema
const postgrestPublic = new PostgrestClient<Database>(REST_URL)

// explicit specification of non-public schema
const postgrestPersonal = new PostgrestClient<Database, 'personal'>(REST_URL, { schema: 'personal' })

postgrestPublic.from('users').select() // select users from the "public" schema

postgresPersonal.from('users').select() // select users from the "personal" schema
```

## What is the new behavior?

A single class instance exposes a facade method to reference other schemas:

```ts
const postgrest = new PostgrestClient<Database>(REST_URL) // implicitly "public" schema

postgrestPublic.from('users').select(); // select users from the "public" schema

postgrestPublic.schema('personal').from('users').select(); // select users from the "personal" schema
```

Type inference for responses also work, assuming a `Database` type has been supplied:
Screenshot 2023-08-05 at 12 09 27 PM

Table schema:
Screenshot 2023-08-05 at 12 10 09 PM

Compare without `.schema('personal')`:
Screenshot 2023-08-05 at 12 13 11 PM

## Additional context

Assuming this PR is accepted, this PR https://github.com/supabase/supabase-js/pull/828 could be modified like so:

```ts
export default class SupabaseClient {
  /* snip */

  /**
   * Perform a query on a schema distinct from the default schema supplied via
   * the `options.db.schema` constructor parameter.
   *
   * The schema needs to be on the list of exposed schemas inside Supabase.
   *
   * @param schema - The name of the schema to query
   */
  schema(
    schema: string & keyof Database
  ): PostgrestClient<Database[typeof schema] extends GenericSchema ? Database[typeof schema] : any, any> {
    return this.rest.schema(schema);
  }
}
```

This would allow the following:

```ts
const supabase = createClient(supabaseURL, supabaseAnonKey);

supabase.schema('personal').from('users').select();
supabase.schema('personal').rpc('get_status', { name_param: 'kiwicopple' });
```

### DIFF
--- a/src/PostgrestClient.ts
+++ b/src/PostgrestClient.ts
@@ -25,7 +25,7 @@ export default class PostgrestClient<
 > {
   url: string
   headers: Record<string, string>
-  schema?: SchemaName
+  schemaName?: SchemaName
   fetch?: Fetch
 
   // TODO: Add back shouldThrowOnError once we figure out the typings
@@ -52,7 +52,7 @@ export default class PostgrestClient<
   ) {
     this.url = url
     this.headers = { ...DEFAULT_HEADERS, ...headers }
-    this.schema = schema
+    this.schemaName = schema
     this.fetch = fetch
   }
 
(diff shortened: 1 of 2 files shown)


soedirgo (approved):
Thanks again!

PR #458, opened 2023-08-08
Make embedded resource non-nullable if using inner join

### LINKED ISSUE #456: !inner should make the relation non-nullable
## Describe the bug

When a relation has the `!inner` hint, the relationship can be considered to always exist due to the embedded resource being included as an inner join. Therefore, `null` can be dropped from the result type union.

## To Reproduce

```ts
const test = supabase().from("people").select("id, site:sites!inner(name)").then(r => r.data);
/* type:
{
    id: string;
    site: {
        name: string | null;
    } | null;
}
*/
```

## Expected behavior

```ts
const test = supabase().from("people").select("id, site:sites!inner(name)").then(r => r.data);
/* type:
{
    id: string;
    site: {
        name: string | null;
    };
}
*/
```

wyozi:

Makes embedded many-to-one resource type non-null when using inner join

Inner joined resources are nullable, even though in practice they must always exist

## What is the new behavior?

Fixes inner joined embedded resource type

## Additional context

fixes #456

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -231,7 +231,9 @@ type ConstructFieldDefinition<
           ? Child | null
           : Relationships extends unknown[]
           ? HasFKey<Field['hint'], Relationships> extends true
-            ? Child | null
+            ? Field extends { inner: true }
+              ? Child
+              : Child | null
             : Child[]
           : Child[]
         : never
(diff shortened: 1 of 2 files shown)


olee:
Hey, it would be cool if this can be merged as we are having the same issue with the incorrect types being generated for embedded, required relations.
Any reason why this does not get merged?

wyozi:
@steve-chavez rebased this one

steve-chavez (approved):

PR #464, opened 2023-08-22
Enforce non-null parameter to .eq

wyozi:

Currently, `.eq()` allows null inputs but crucially it won't work, as you need to use `.is()` for null comparisons.

This PR adds stronger typing to ensure that null values are not passed to `.eq()`. Instead, you'll either have to do non-null assertion or a more proper method, like `.is()` or `.eqNullable()` (#463)

Note: this is kind of a breaking change, but I think an extremely useful one and call sites where this causes errors should be fixed


soedirgo (approved):

soedirgo:
Thanks!

> this is kind of a breaking change, but I think an extremely useful one and call sites where this causes errors should be fixed

Agreed, using `null` with `.eq()` would just fail in the first place.

PR #486, opened 2023-10-11
fix: detection of maybeSingle

### LINKED ISSUE #361: Error logged when using `.maybeSingle()` returns no result

steve-chavez:
Fixes #361.

On PostgREST 11.2.0, the format of the error got changed on PostgREST/postgrest#2876 to "The result contains 0 rows".

Change the detection to "0 rows". This should be backwards compatible.

### DIFF
--- a/src/PostgrestBuilder.ts
+++ b/src/PostgrestBuilder.ts
@@ -149,7 +149,7 @@ export default abstract class PostgrestBuilder<Result>
           }
         }
 
-        if (error && this.isMaybeSingle && error?.details?.includes('Results contain 0 rows')) {
+        if (error && this.isMaybeSingle && error?.details?.includes('0 rows')) {
           error = null
           status = 200
           statusText = 'OK'


src/PostgrestBuilder.ts
```diff
@@ -149,7 +149,7 @@ export default abstract class PostgrestBuilder<Result>
           }
         }
 
-        if (error && this.isMaybeSingle && error?.details?.includes('Results contain 0 rows')) {
+        if (error && this.isMaybeSingle && error?.details?.includes('0 rows')) {
```
kamilogorek, line 152: Is there a reason why we can't rely on error codes instead? Is this the fact that `ApiRequestErrorCode16` is an umbrella error for "incorrect number of rows", and not just 0?

soedirgo: Yeah, we need to distinguish 0 rows from >1 rows

soedirgo (approved):

src/PostgrestBuilder.ts
soedirgo: Still not a fan of it though - ideally we can check this in a more robust manner

PR #491, opened 2023-10-31
feat: detect one-to-one relationships

soedirgo:
This can be enabled by supplying an extra type argument:

```ts
const postgrest = new PostgrestClient<Database, 'public', { detectOneToOneRelationships: true }>(...)
```

feature

Doesn't detect one-to-one relationships - foreign table would be typed as `T[]`

## What is the new behavior?

Foreign table on a one-to-one relationship would be typed as `T | null`


thorwebdev (approved):
LGTM

PR #493, opened 2023-11-02
chore: `foreignTable` -> `referencedTable`

soedirgo:

docs update

We use the term "foreign table" for referenced tables but [Postgres already use that term for something else](https://www.postgresql.org/docs/current/sql-createforeigntable.html).

## What is the new behavior?

Deprecate `foreignTable` and use `referencedTable` moving forward


dshukertjr (approved):
LGTM!

PR #506, opened 2023-11-17
fix: explain() always returning an error

### LINKED ISSUE #504: .explain() method always return error
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Chaining the .explain() method to a working query, return this error:
``` 
error: {
    code: '42883',
    details: null,
    hint: 'No function matches the given name and argument types. You might need to add explicit type casts.',
    message: 'function string_agg(record, unknown) does not exist'
  } 
```

## To Reproduce
Following the official guide:
https://supabase.com/docs/guides/api/rest/debugging-performance

Run this SQL command:
``` 
alter role authenticator set pgrst.db_plan_enabled to 'true';
NOTIFY pgrst, 'reload config'; 
```

Query some data:
``` 
const data = await supabase.from('countries').select('*').explain()
console.log(data)
```

## Expected behavior

Response similar to this:
``` 
Aggregate  (cost=33.34..33.36 rows=1 width=112)
  ->  Limit  (cost=0.00..18.33 rows=1000 width=40)
        ->  Seq Scan on countries  (cost=0.00..22.00 rows=1200 width=40) 
```

## Additional context

I've tried with /supabase-js & /ssr

laurenceisla:

Bug fix. Closes #504

Sends `"undefined"` media type when requesting explain plan, making the explain() method fail.

## What is the new behavior?

Sends `"application/json"` as media type.

## Additional context

Tests weren't failing because they were using PostgREST v11.0.0 instead of v11.2.2

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -267,7 +267,7 @@ export default class PostgrestTransformBuilder<
       .filter(Boolean)
       .join('|')
     // An Accept header can carry multiple media types but postgrest-js always sends one
-    const forMediatype = this.headers['Accept']
+    const forMediatype = this.headers['Accept'] ?? 'application/json'
     this.headers[
       'Accept'
     ] = `application/vnd.pgrst.plan+${format}; for="${forMediatype}"; options=${options};`
--- a/test/db/docker-compose.yml
+++ b/test/db/docker-compose.yml
@@ -3,7 +3,7 @@
 version: '3'
 services:
   rest:
-    image: postgrest/postgrest:v11.0.0
+    image: postgrest/postgrest:v11.2.2
     ports:
       - '3000:3000'
     environment:
--- a/test/transforms.ts
+++ b/test/transforms.ts
@@ -200,7 +200,7 @@ test('maybeSingle', async () => {
       "data": null,
       "error": Object {
         "code": "PGRST116",
-        "details": "Results contain 2 rows, application/vnd.pgrst.object+json requires 1 row",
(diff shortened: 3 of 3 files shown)


src/PostgrestTransformBuilder.ts
```diff
@@ -267,7 +267,7 @@ export default class PostgrestTransformBuilder<
       .filter(Boolean)
       .join('|')
     // An Accept header can carry multiple media types but postgrest-js always sends one
-    const forMediatype = this.headers['Accept']
+    const forMediatype = 'application/json'
```
steve-chavez, line 270: @laurenceisla There's also `csv` and `geojson()` transforms. Would `explain()` work with those too?

steve-chavez: https://supabase.com/docs/reference/javascript/db-csv

steve-chavez: Maybe the fix is to assign application/json whenever accept is not defined

laurenceisla: Yes, a conditional works nicely. A test was missing for these transformations, so I'll add it.

laurenceisla: nvm... it needs new changes to make explain and csv/geojson to work together in TypeScript. I added the conditional in the meantime.

steve-chavez (approved):

PR #507, opened 2023-11-21
fix: add link on how to enable explain on the explain() method

dshukertjr:

Thought it would be helpful to include a link on how to enable the `explain()` feature on the method itself so that people know how to enable it.

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -224,6 +224,12 @@ export default class PostgrestTransformBuilder<
   /**
    * Return `data` as the EXPLAIN plan for the query.
    *
+   * Before using this method, you need to enable `explain()` on your
+   * Supabase instance by following the guide below. Note that `explain()`
+   * should only be enabled on an development environment.
+   *
+   * https://supabase.com/docs/guides/api/rest/debugging-performance#enabling-explain
+   *
    * @param options - Named parameters
    *
    * @param options.analyze - If `true`, the query will be executed and the
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -224,11 +224,7 @@ export default class PostgrestTransformBuilder<
   /**
    * Return `data` as the EXPLAIN plan for the query.
    *
-   * Before using this method, you need to enable `explain()` on your
-   * Supabase instance by following the guide below. Note that `explain()`
-   * should only be enabled on an development environment.
-   *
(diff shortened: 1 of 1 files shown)


steve-chavez (approved):

soedirgo:
Let's put this in Supabase docs instead - it's a bit odd to have Supabase-specific instructions in postgrest-js

dshukertjr:
@soedirgo 
Sure, but do people use this library out side of Supabase's context though? Or do we have any plans of promoting postgrest-js to the general PostgREST community? 

I personally would find it helpful if the IDE told me that this method doesn't work unless I take specific actions when I hover over it. 

Screenshot 2023-11-28 at 17 30 05

soedirgo:
We do promote it as a standalone library: https://postgrest.org/en/stable/ecosystem.html#client-side-libraries

While there's merit to only sanctioning the hosted platform (less environments to maintain), we've committed to making sure each service works on its own, and this runs counter to that. E.g. this wouldn't help people using the library on a self-hosted setup (they may need to update the Compose file instead), or local development via Supabase CLI (they may need to update `supabase/config.toml`).

But I agree, the tradeoff is that the advice is not front and center. Alternatively we can update the docs page to show how to enable `.explain()` on all environments.

dshukertjr:
Sounds good. I will close this one!

soedirgo:
Sorry, I sent that too quick 😬 Since the docs page is already not Supabase specific, we can change the language to make it not Supabase-specific, e.g. "you need to set pgrst.db_plan_enabled to 'true'", since the steps in the docs page also applies for a generic PostgREST setup

dshukertjr:
That makes sense. Let me update it real quick now

src/PostgrestTransformBuilder.ts
```diff
@@ -224,6 +224,12 @@ export default class PostgrestTransformBuilder<
   /**
    * Return `data` as the EXPLAIN plan for the query.
    *
+   * Before using this method, you need to enable `explain()` on your
+   * Supabase instance by following the guide below. Note that `explain()`
+   * should only be enabled on an development environment.
+   *
+   * https://supabase.com/docs/guides/api/rest/debugging-performance#enabling-explain
+   *
```
dshukertjr, line 232:
```suggestion
   * You need to set `pgrst.db_plan_enabled` to `true` before using this method.
   *
```

dshukertjr: @soedirgo Something like this?

soedirgo: Yep, thanks!

PR #527, opened 2024-03-23
Allow to pass Next.js options to fetch while building query

g12i:

This adds a possibility to pass `next` options to [fetch](https://nextjs.org/docs/app/api-reference/functions/fetch#optionsnextrevalidate) while building query.

Arguably there could be possibility to overwrite _any_ `RequestInit` options, e.g. [`cf`](https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties). There would have to be some strategy to merge these with defaults, but it could offer more flexibility. 

Currently we can pass fetch at `PostgrestBuilder` creation. Which effectively means once, if using Supabase. One would have to parse URL and build tags in central place.

```ts
  const postgrest2 = new PostgrestClient<Database>('http://localhost:3000', {
    fetch: (url, init) => {
      let parsedUrl: URL

      if (url instanceof URL) {
        parsedUrl = url
      } else if (typeof url === 'string') {
        parsedUrl = new URL(url)
      } else {
        parsedUrl = new URL(url.url)
      }

      const tags: string[] = []
      if (parsedUrl.pathname.startsWith('/rest')) {
        const [,, entity] = parsedUrl.pathname.split('/').slice(1)
        tags.push(entity)
      }

      return fetch(url, {
        ...init,
        next: {
          tags,
        },
      })
    },
  })

```

## What is the new behavior?

Allow to pass `next` options when building query. The idea here is 

```ts
  const res = await postgrest
    .from('users')
    .select()
    .eq('username', 'supabot')
    .next({
      tags: ['users/supabot'],
    })

// ... later in the code

revalidateTag('users/supabot');
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -70,3 +70,18 @@ export type GenericSchema = {
 
 // https://twitter.com/mattpocockuk/status/1622730173446557697
 export type Prettify<T> = { [K in keyof T]: T[K] } & {}
+
+export type NextFetchRequestConfig = RequestInit extends { next: infer T }
+  ? T
+  : {
+      revalidate?: number | false
+      tags?: string[]
+    }
+
+declare global {
+  namespace globalThis {
+    interface RequestInit {
+      next?: unknown // Avoid: 'next' is referenced directly or indirectly in its own type annotation.ts(2502)
+    }
+  }
+}
--- a/src/PostgrestBuilder.ts
+++ b/src/PostgrestBuilder.ts
@@ -1,7 +1,7 @@
 // @ts-ignore
 import nodeFetch from '@supabase/node-fetch'
 
-import type { Fetch, PostgrestSingleResponse } from './types'
+import type { Fetch, NextFetchRequestConfig, PostgrestSingleResponse } from './types'
 import PostgrestError from './PostgrestError'
 
 export default abstract class PostgrestBuilder<Result>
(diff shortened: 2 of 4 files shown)


src/types.ts
```diff
@@ -70,3 +70,18 @@ export type GenericSchema = {
...
+      tags?: string[]
+    }
+
+declare global {
+  namespace globalThis {
+    interface RequestInit {
+      next?: unknown // Avoid: 'next' is referenced directly or indirectly in its own type annotation.ts(2502)
+    }
+  }
+}
```
g12i, line 87: I'm not entirely happy with this part. The idea is not to overwrite `NextFetchRequestConfig` if someone is using this in Next.js environment. However later I'm overwriting global RequestInit. Probably there is a better way to do this.

bnjmnt4n:
Instead of this being Next.js specific, not sure if it might be better to just allow customization of any `fetch` option?

g12i:
Yeah, I agree @ bnjmnt4n. Let me make a quick update

g12i:
Hello @soedirgo, any chance this gets reviewed?

toniab:
@soedirgo bump :P

aiddun:
Any update on this?

mandarini:
Hi @g12i ! Thank you very much for you meaningful contribution, and apologies for not reviewing it for such a long time.

This repository is deprecated and has moved to the new [Supabase JS monorepo](ttps://github.com/supabase/supabase-js). 

May I ask you, if you believe this change is still needed, please open a new PR in the monorepo and include a link back to this thread for context:

I will be closing this PR now, but you can always access it and view the changes, even after the repo is archived, don't worry!

PR #530, opened 2024-04-03
detect !inner in one-to-one relationships

wyozi:

Select query type parser feature/bugfix

```ts
.select("a_onetoone_rel!inner(*)")
```

results in a nullable `a_onetoone_rel`, while it should be non-nullable due to `!inner`

## What is the new behavior?

Detects this case and removes nullability of the inner one-to-one relationship.

## Additional context

non one-to-one case was fixed in #458

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -228,7 +228,9 @@ type ConstructFieldDefinition<
               ? R
               : unknown
           > extends true
-          ? Child | null
+          ? Field extends { inner: true }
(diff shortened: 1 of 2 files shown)


wyozi:
@steve-chavez this would be good to have merged, since otherwise https://github.com/supabase/postgrest-js/pull/458 will behave very unexpectedly with `!inner` randomly not working on one-on-one relationships

soedirgo (approved):

Issue #524, opened 2024-03-11
[supabase-js] Parser error for using aggregate and spread


thecoorum:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create a typed Supabase client with (types are generated with `npx supabase gen types typescript --local`
   ```
   createClient<Database>(...)
   ```
2. Try to query data
  2a. Using spread in `select`
        ```
        .select('...users (first_name, last_name)')
        ```
 2b. Using newly introduced aggregate functions
        ```
        .select('from:from_time.min(), till:till_time.max()')
        ```

2a will type fail with `PostgrestSingleResponse<ParserError<"Expected identifier at '...users (first_name, last_name'">>`
2b will type fail with `PostgrestSingleResponse<ParserError<"Unexpected input: .min(), till:till_time.max()">>`

## Expected behavior

No TypeScript errors

## System information

- OS: macOS
- Version of supabase-js: `^2.39.7`
- Version of Node.js: `19.4.0`

stefan-girlich:
@encima Would you know if there is a workaround? Right now, I'm casting all outputs to `any` before further processing. Maybe there is an casting-based approach utilizing the Supabase types that are already generated.

bnjmnt4n:
@stefan-girlich are you using the latest version of supabase-js?

stefan-girlich:
Thanks for the hint. I upgraded supabase-js, but I'm now seeing a new error. I added an updated in my [original bug report](https://github.com/supabase/postgrest-js/issues/523#issuecomment-2047393866).

avallete:
Hey everyone !

We've reworked the result inference in `postgrest-js` to address the typing issues mentioned above. The changes are available in a canary release, and we're actively seeking your feedback.

To test it out, update your `supabase-js` to version `2.46.0-rc.1` (which includes `postgrest-js` `1.17.0-rc.1`):

```bash
npm install supabase-js@2.46.0-rc.1
```

Please let us know if you encounter any issues, especially with retro-compatibility, so we can finalize it for the next release.

avallete:
[...]

Issue #523, opened 2024-03-11
ParserError "Unexpected input: .sum()" when using aggregate functions with TypeScript


stefan-girlich:
# Bug report
- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When I use aggregate functions, the inferred TypeScript data type will always depict an error, requiring me to use casts to `any`.

## To Reproduce

```ts
  const { data } = await supabase.from("things").select(`
    field.sum(),
    ...other_things(id, name, color)
  `);
```

The inferred type for `data` seen in my VS Code is:
```ParserError<"Unexpected input: .sum(),\n    ...other_things(id, name, color)\n  ">[] | null```

Inference works fine when not using aggregate functions.
Re-generating TS types via `supabase gen types typescript` does not fix the issue.

## Expected behavior
Types should be inferred correctly without an error.

## System information

- OS: MacOS
- Version of supabase-js: 2.39.7
- Version of Node.js: 21.6.2

**Comment by stefan-girlich** [comment]
A brief update after upgrading `supabase-js` from 2.39.7 to 2.42.0:

Error in TypeScript type is now as following for the same original code:
```ts
const data: SelectQueryError<"Referencing missing column `sum`">[] | null
```

`@supabase/postgrest-js` is at v1.15.0, so the fix for #447 should already be included.
Am I maybe missing something?

stefan-girlich:
@bnjmnt4n I added [test cases](https://github.com/supabase/postgrest-js/commit/8ba7b3e846f24d9c07a3dfb7224077acce379adc) which currently fail at compile-time.

bnjmnt4n:
@stefan-girlich thanks for providing the examples; I'm really busy now so I can probably look at them next week only.

stefan-girlich:
@bnjmnt4n Would you know if this is scheduled already? Thanks!

stefan-girlich:
@bnjmnt4n Would you know more? I can look into a solution/PR myself, but I'd be happy about some guidance concerning how to tackle this.

sanjaybaskaran01:
Is this resolved? Facing the same issue here.

veloware:
@sanjaybaskaran01 dont think it has been resolved, im also seeing the same when trying to use `sum()`

[...]

Issue #520, opened 2024-03-06
Typescript Support for Aggregate Functions


ConProgramming:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Aggregate functions, just added in #206 are not supported in Typescript.

So for example will get the error ParserError<"Expected identifier at ), label">[].

Would love to see fixed quick like #459 was.

bnjmnt4n:
I can send a PR for this.

PR #531, opened 2024-04-10
#523 fix typescript error for aggregate functions

stefan-girlich:

Fix dynamic TypeScript type for `data` when using aggregate functions other than `count`

Aggregate functions other than `count` will result in TypeScript errors for the resulting type of `data` (#523)

Bug fix for issues:
- #523 
- #524 (potentially, not tested explicitly)

## What is the new behavior?

Using the other aggregate functions "sum", "avg", "min", and "max" will not result in an erroneous TypeScript type for `data`.

## Additional context

I just generalized changes from #520 by @bnjmnt4n ; thanks for your work!

### DIFF
--- a/src/select-query-parser.ts
+++ b/src/select-query-parser.ts
@@ -274,8 +274,8 @@ type ConstructFieldDefinition<
   : Field extends { name: string; original: string }
   ? Field['original'] extends keyof Row
     ? { [K in Field['name']]: Row[Field['original']] }
-    : Field['original'] extends 'count'
-    ? { count: number }
+    : Field['original'] extends AggregateFunctions
+    ? { [K in Field['original']]: number }
     : SelectQueryError<`Referencing missing column \`${Field['original']}\``>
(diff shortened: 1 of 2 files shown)


stefan-girlich:
@bnjmnt4n You're right; `column(max)` is not valid; my fix is therefore not ideal as it allows for invalid syntax.
Would you have any idea concerning how to fix the issues with the other aggregation functions besides `count()` (#523 , #524 )?

bnjmnt4n:
Do you have a reproducible failing test case or example?

stefan-girlich:
Sure; I mentioning them in #523  for visibility: https://github.com/supabase/postgrest-js/issues/523#issuecomment-2048080742

avallete:
A global fix has now been released in `supabase-js` v2.46.0, covering this issue. Thank you for your effort and work on this PR—it’s much appreciated! I’ll be closing this PR, but please feel free to reach out if there are any further issues.

PR #537, opened 2024-05-21
fix: use proper dual-packaging

soedirgo:
Support CommonJS & ESM. Previously the ESM build (dist/module) is broken - it's recognized as CommonJS.

For more on dual-packaging: https://nodejs.org/api/packages.html#dual-commonjses-module-packages


thorwebdev (approved):
LGTM, thanks!

PR #540, opened 2024-05-27
Reland ESM fix

lucacasonato:
Relanding https://github.com/supabase/postgrest-js/pull/540/commits/6ac04ffd404265230d4aa2c262970ce976ff8ab1, but fixed to support this usecase:

```js
import * as ns from "@supabase/postgrest-js"

new ns.PostgrestClient()
```


soedirgo (approved):
Thanks! Works on Supabase Edge Functions 💯

PR #548, opened 2024-07-04
fix: "Cannot destructure property 'PostgrestClient'" error

itsjohnward:

Bug fix

Getting an error here:

```
Cannot destructure property 'PostgrestClient' of '_index.default' as it is undefined.
```

because index.ts is only exporting named exports with no default export.

## What is the new behavior?

Import works as expected, and I can create a SupabaseClient

### DIFF
--- a/wrapper.mjs
+++ b/wrapper.mjs
@@ -1,11 +1,10 @@
-import index from '../cjs/index.js'
-const {
+import {
   PostgrestClient,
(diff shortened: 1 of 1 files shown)


soedirgo:
Thanks for the PR! This should be fixed in https://github.com/supabase/postgrest-js/pull/549, so closing this.

PR #550, opened 2024-07-10
Update PostgrestClient.ts to allow for additional headers on-demand

activenode:

Feature

Currently you cannot use the postgrest-js to add additional headers which however is required in use-cases such as described in https://supabase.com/docs/guides/api/securing-your-api?queryGroups=database-method&database-method=sql&queryGroups=pre-request&pre-request=rate-limit-per-ip 

## What is the new behavior?

One can pass `additionalHeaders` and does not have to build a custom wrapper.

### DIFF
--- a/src/PostgrestBuilder.ts
+++ b/src/PostgrestBuilder.ts
@@ -47,6 +47,15 @@ export default abstract class PostgrestBuilder<Result>
     return this
   }
 
+  /**
+   * Set an HTTP header for the request.
+   */
+  setHeader(name: string, value: string): this {
+    this.headers = { ...this.headers }
+    this.headers[name] = value
+    return this
+  }
+
   then<TResult1 = PostgrestSingleResponse<Result>, TResult2 = never>(
     onfulfilled?:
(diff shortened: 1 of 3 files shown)


soedirgo:
Thanks! Does the `headers` option in the `PostgrestClient` constructor not work for this?

activenode:
The problem is that it's global. But headers are dynamic. Mapped back to the sample provided it would mean that, for each change, I'd have to instantiate a new Supabase-js client (using the underlying Postgres client)

soedirgo:
Makes sense, thanks! How about a `setHeader()` method in `PostgrestBuilder`? That allows you to use it more generically, e.g. with `.from()`.

activenode:
that sounds great

activenode:
shall I go for it?

soedirgo:
Hey! Went ahead and updated the PR - if you can verify this works for your use case I'll go ahead and merge this 👍

activenode:
Wow, great! Will do.

activenode:
image
tested both rpc and select, works damn fine.

soedirgo (approved):
Awesome! Merging now 🚀

Issue #546, opened 2024-06-17
Inconsistent Typing for Joined Field in Supabase


cptcrunchy:
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I have added a link to a REPL, [HERE](https://replit.com/@cptcrunchy/SelfreliantObviousPlan?v=1) along with a screenshot.
The example code expects the `created_by` field in the query result to be a single object containing employee details (id and full_name). However, the type definition `({ id: any; full_name: any; }[])` suggests it's an array of employees. This mismatch causes a typing error when working with the data.

**Why it happens**:

The !report_created_by_fkey syntax likely defines a relationship between tables, but the return type might not be explicitly set to a single object.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to The REPL [link](https://replit.com/@cptcrunchy/SelfreliantObviousPlan?v=1) 
2. Enter your Project URL and ANON key into the createClient instance.
3. Create the table schemas provided in `$lib/db`
4. When hovering over `data` the return type def is an array of objects.

## Expected behavior

The result type def should be a single object for the field `created_by`.

## Screenshots

![Screenshot 2024-06-17 at 15 57 57](https://github.com/supabase/supabase/assets/8497282/a751bf0a-66a8-4d36-ac3f-e75cf7c02c88)

## System information

- OS: macOS
- Version of supabase-js: 2.42.7
- Version of Node.js: 18.20.3

jdgamble555:
@soedirgo - This is the same issue opened MANY times and still needs to be fixed.
- https://github.com/supabase/postgrest-js/issues/512
- https://github.com/supabase/postgrest-js/issues/536
- https://github.com/supabase/postgrest-js/issues/509
- https://github.com/supabase/postgrest-js/issues/471
- https://github.com/supabase/postgrest-js/issues/434
- https://github.com/supabase/postgrest-js/issues/408
- https://github.com/supabase/postgrest-js/issues/406
- https://github.com/supabase/postgrest-js/issues/355

We have 8 different repos all for the same problem. JOINS do not have the correct typing. This needs to be moved up the ladder as a PRIORITY.

J

jdgamble555:
[...]

PR #553, opened 2024-09-08
fix(select-query-parser): select left join and array instead of objects result types

avallete:
This PR improves the TypeScript typings generated by the SQL query parser, fixing issues with inferred types for certain `!left` joins and relationship queries (e.g., `created_by:users(*)`).

Currently, for certain queries, the inferred types for relations are incorrect. For instance, the type may be inferred as `{}[]` when the runtime result is a single object `{}`.

**Related issues**
https://github.com/supabase/postgrest-js/issues/546
https://github.com/supabase/supabase/issues/29086

## What is the new behavior?
This PR updates the select-query-parser to handle these cases more accurately by:

- Extending the handling logic for the `!left` operator, similar to the existing logic for the `!inner` operator.
- Returning a single object for relationships when the runtime result is also a single object.

## Additional context
The TypeScript code in select-query-parser is quite complex. I would appreciate a thorough review of this fix.
Additional tests have been added to cover both runtime and typing scenarios. This helped uncover more edge cases, some of which still need further solutions.

### Known Issues:
Currently, there is no distinction between possibly nullable and non-nullable results for zero-to-one or one-to-one relations. To handle this properly, the PR always returns null as a possible option. However, a more accurate solution would involve enhancements like adding an `isOneToOne` option in Relationships that marks foreign keys as nullable or non-nullable based on `columns[]` type definitions.

### PS
Also took the chance to fix the CI workflow by replacing the [deprecated docker-compose](https://github.blog/changelog/2024-04-10-github-hosted-runner-images-deprecation-notice-docker-compose-v1/) with `docker compose` command.


test/db/01-dummy-data.sql
```diff
@@ -30,3 +30,18 @@ VALUES
...
+
+INSERT INTO user_profiles (id, username)
+VALUES
+  (1, 'supabot'),
+  (2, NULL);
+
+INSERT INTO best_friends(id, first_user, second_user, third_wheel)
+VALUES
+  (1, 'supabot', 'kiwicopple', 'awailas'),
+  (2, 'supabot', 'awailas', NULL);
```
soedirgo, line 47:
```suggestion
  (2, 'supabot', 'awailas', NULL);

```

soedirgo:
Thanks! I think this fixes https://github.com/supabase/supabase/issues/29086 but not https://github.com/supabase/postgrest-js/issues/546 - will investigate that separately

soedirgo (approved):

PR #555, opened 2024-09-27
Export PostgresError

blechatellier:
Exporting the custom `PostgrestError` error, not just the type, so we can check `instanceof PostgresError` in code.


soedirgo (approved):

soedirgo:
Thanks! 🙏

PR #557, opened 2024-09-27
chore: increase runtime tests coverage

avallete:

A tests cases to assert postgrest runtime behavior, their values are simply here to act as a canary across versions upgrades.

They will be reused to ensure better coverage of the type result inference.

## What is the new behavior?

## Additional context


soedirgo:
I think it'd be more useful to move these into type tests (`index.test-d.ts`) - we're just passing these strings to PostgREST so there wouldn't be a lot of postgrest-js bugs that can be caught by runtime tests. We do cover basic tests for embedded tables in `resource-embedding.ts`.

soedirgo (approved):

PR #554, opened 2024-09-11
fix: query result parser

avallete:
## What kind of change does this PR introduce?

wip: an attempt to solve #546 and add more test coverage for consistency between runtime and raised types

## What is the current behavior?

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

Issue #450, opened 2023-07-26
Types for column name disambiguation


wyozi:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

PostgREST allows disambiguating relationships by using a column name as the target directly (see column name section of https://postgrest.org/en/stable/references/api/resource_embedding.html#embedding-disambiguation). However, select query parser in postgrest-js does not support this yet.

Personally I find column name disambiguation to be the superior method for embeds, so it would be nice to have this

## To Reproduce

Use column name instead of foreign key constraint name or relation name as the target, i.e. `.select('user:username(*)')` instead of `.select('user:users(*)')`

albinekb:
I have been using the patch from #451 for three months now, it's working great! Any chance of getting that merged? It's a bit messy to have to manage the patch. Thanks!

avallete:
Hey everyone !

We've reworked the result inference in `postgrest-js` to address the typing issues mentioned above. The changes are available in a canary release, and we're actively seeking your feedback.

To test it out, update your `supabase-js` to version `2.46.0-rc.1` (which includes `postgrest-js` `1.17.0-rc.1`):

```bash
npm install supabase-js@2.46.0-rc.1
```

Please let us know if you encounter any issues, especially with retro-compatibility, so we can finalize it for the next release.

wyozi:
I'm using an extended database types definition (something similar to https://github.com/activenode/supabase-table-extensions-typed) and it seems to break with this error:

```
      Property 'supabaseUrl' is protected but type 'SupabaseClient<Database, SchemaName, Schema>' is not a class derived from 'SupabaseClient<Database, SchemaName, Schema>'.
```

wyozi:
looks like there's a small regression with `!inner`; in some situations it doesn't make the relation non-nullable but in some situations it does. I'll try to research it a bit

avallete:
Hey @wyozi 

Thank's for giving it a try and sending feedbacks ! Happy to hear that it's working for you and you managed to get everything working on the new version !

avallete:
The fix has been released in `supabase-js` v2.46.0. I'm closing this issue, but feel free to reopen or open a new issue if you encounter any further errors.

Issue #408, opened 2023-03-05
Database typings are not correct on table join for entity to one relationship


niels-bosman:
# Bug report

## Describe the bug

Typings are not correct when joining tables that are always NOT an array. For example, in my application we use the following query to fetch data about `figma_comments` and link to `figma_users` and `figma_files`:

```ts
const { data: comments } = await supabase
    .from('figma_comments')
    .select(
      'created_at, resolved_at, comment, node_id, user:figma_users(handle, img_url, email), file:figma_files(file_key, name)',
    )
    .order('created_at', { ascending: false });
```

From this query, I would expect the built in typings to have both `user` and `file` to be an entity of a singular user and a singular file. Since the constraints I set up in Postgres won’t allow it to be multiple entries because the foreign key is unique on the joinable tables.

Here is an overview of the types being generated by supabase CLI:

```ts
Tables: {
      figma_comments: {
        Row: {
          comment: string;
          created_at: string;
          figma_user_id: string;
          file_key: string;
          id: number;
          node_id: string;
          order_id: number | null;
          parent_id: string | null;
          resolved_at: string | null;
        };
        Insert: {
          comment: string;
          created_at?: string;
          figma_user_id: string;
          file_key: string;
          id?: number;
          node_id: string;
          order_id?: number | null;
          parent_id?: string | null;
          resolved_at?: string | null;
        };
        Update: {
          comment?: string;
          created_at?: string;
          figma_user_id?: string;
          file_key?: string;
          id?: number;
          node_id?: string;
          order_id?: number | null;
          parent_id?: string | null;
          resolved_at?: string | null;
        };
      };
      figma_files: {
        Row: {
          editor_type: string | null;
          figma_project_id: number | null;
          file_key: string;
          id: number;
          last_modified: string;
          name: string;
          thumbnail_url: string;
          version: string | null;
        };
        Insert: {
          editor_type?: string | null;
          figma_project_id?: number | null;
          file_key: string;
          id?: number;
          last_modified: string;
          name: string;
          thumbnail_url: string;
          version?: string | null;
        };
        Update: {
          editor_type?: string | null;
          figma_project_id?: number | null;
          file_key?: string;
          id?: number;
          last_modified?: string;
          name?: string;
          thumbnail_url?: string;
          version?: string | null;
        };
      };
      figma_users: {
        Row: {
          email: string;
          figma_id: string;
          handle: string;
          id: number;
          img_url: string;
        };
        Insert: {
          email: string;
[...]

PR #558, opened 2024-09-27
feat: select-query-parser-v2

avallete:

- Introduces new, separate code for the parser and result types, each tested independently.
- Adds corresponding type tests for each runtime test case.
- Fixes several bugs related to type result inference.

There are several result type mismatches referenced across multiple issues:

- https://github.com/supabase/postgrest-js/issues/546
- https://github.com/supabase/postgrest-js/issues/524
- https://github.com/supabase/postgrest-js/issues/509
- https://github.com/supabase/postgrest-js/issues/450
- https://github.com/supabase/postgrest-js/issues/408

I have reproduced these issues in test cases with similar topologies to verify runtime behavior and ensure proper type inference.

## What is the new behavior?

The rewrite should resolve all the related issues regarding select operations that I’ve encountered so far. However, I may have missed some cases, and there could be regressions in edge cases (despite passing all existing tests). Given the number of related issues, I’m unsure if the previous test coverage was sufficient. I recommend to deploy that out over an `rc` version first in coordination with `supabase-js rc` candidate so we can gather feedbacks about possibles regression before making the switch.

Additionally, we might need to evaluate how this performs on large databases with many tables and relations. The use of recursive types could potentially introduce performance concerns in such cases.

## Additional context

I didn’t rewrite the original parser for fun. After spending a considerable amount of time trying to fix the existing one (I swear I tried see: https://github.com/supabase/postgrest-js/pull/554 😓 ), it became clear that starting fresh with a cleaner, more testable, and debuggable version was necessary. I know it make for a big PR but I wasn't able to really reduce more than that the amount of changes, sorry 😅


src/PostgrestClient.ts
```diff
@@ -21,7 +21,8 @@ export default class PostgrestClient<
     : string & keyof Database,
   Schema extends GenericSchema = Database[SchemaName] extends GenericSchema
     ? Database[SchemaName]
-    : any
+    : any,
+  TypesVersion extends 'next' | undefined = undefined
```
avallete, line 25: Here is our feature flag, it needs to be drilled down to `PostgrestQueryBuilder` and  `Filter` but that's the best way I've found to implement it. Open to suggestions tough.

src/types.ts
```diff
@@ -39,20 +39,31 @@ export type PostgrestSingleResponse<T> = PostgrestResponseSuccess<T> | Postgrest
...
   Row: Record<string, unknown>
   Insert: Record<string, unknown>
   Update: Record<string, unknown>
+  Relationships: GenericRelationship[]
 }
 
 export type GenericNonUpdatableView = {
   Row: Record<string, unknown>
+  Relationships: GenericRelationship[]
 }
```
avallete, line 67: Looking at the types all of those have relationships defined even if empty. This help with dealing with more similar types all around the codebase.

test/index.test-d.ts
```diff
@@ -80,15 +79,6 @@ const postgrest = new PostgrestClient<Database>(REST_URL)
...
 
-// embedded resource with no fields
-{
-  const { data, error } = await postgrest.from('messages').select('message, users()').single()
-  if (error) {
-    throw new Error(error.message)
-  }
-  expectType<{ message: string | null }>(data)
-}
-
```
avallete, line 91: Moved all those tests into `select.test-d.ts`

test/select-query-parser/select.test-d.ts
avallete, line 1: You can check it all but it's basically the type result implementation of each matching runtime test in `relation.ts`

serranoarevalo:
Will this also fix: https://github.com/supabase/supabase-js/issues/930 ?

When using something like:

```
review_avg:reviews!product_id (
    rating.avg()
),
```

The generated type is:

```typescript
{
    review_avg: SelectQueryError<"Referencing missing column `avg`">[];
}

**Comment by avallete** [comment]
> Will this also fix: [supabase/supabase-js#930](https://github.com/supabase/supabase-js/issues/930) ?
> 
> When using something like:
> 
> ```
> review_avg:reviews!product_id (
>     rating.avg()
> ),
> ```
> 
> The generated type is:
> 
> ```ts
> {
>     review_avg: SelectQueryError<"Referencing missing column `avg`">[];
> }
> ```

I should even if that's hard to tell without the exact database types along with the query but there is now some tests accounting for "column aggregation" similar to what you got here:

https://github.com/supabase/postgrest-js/blob/49ba6fd637c0d60b734325d7900483f8157636a8/test/relationships.ts#L125-L142

Giving expected results:

https://github.com/supabase/postgrest-js/blob/49ba6fd637c0d60b734325d7900483f8157636a8/test/select-query-parser/select.test-d.ts#L535-L566

**Comment by avallete** [comment]
Just tried to put this new version on a big project with lot of calls to the postgrest client. It have some errors that I need to fix, so the review on this can wait.

File: src/select-query-parser/utils.ts [hunk 4]
```diff
@@ -0,0 +1,495 @@
...
+ */
+export type DeduplicateRelationships<T extends readonly unknown[]> = T extends readonly [
+  infer First,
+  ...infer Rest
+]
+  ? First extends Rest[number]
+    ? DeduplicateRelationships<Rest extends readonly unknown[] ? Rest : []>
+    : [First, ...DeduplicateRelationships<Rest extends readonly unknown[] ? Rest : []>]
+  : T
+
```
avallete, line 55: **remark**

This should probably be actually handled at `pg-meta` level to avoid duplicating relationships that have the exact same content. Either by providing the `referencedSchema` value or deduplicating the actual keys.

But I think this is actually a much deeper issue about how we handle things, right now everything is focused over a single "schema" so "cross-schema" relations won't work type-wise. I'm not sure but I think it would work at postgrest runtime tough.

I'm pretty confident that this could be implemented at type level as well tough. But it would require changes over the generated `pg-meta` types. So for now, I decided to go to the path of less frictions by simply deduping the relationships to avoid "required hinting errors" to be prompted in such case. But still have types warning where postgrest will actually ask for disambiguation at runtime.

whollacsek:
Would this PR fix type error when using virtual columns (ie. calling postgres function as column)?

avallete:
> Would this PR fix type error when using virtual columns (ie. calling postgres function as column)?

Hey, I'm not sure what issue you're referring too I don't think so, from my understanding the only way to call functions in postgrest is via RPC (https://docs.postgrest.org/en/v12/references/api/functions.html).

The only way I think you could achieve to use them as columns would be via some kind of views maybe ?

Also note that this PR does not impact any runtime behavior, only make the result types inference match it better.

whollacsek:
What I was referring to does work, my question is about fixing the generated type as the typescript error message is preventing builds of my next.js app to succeed. I've been forced to add ts-ignore everywhere since 2 years and I was thinking that this PR sounds like it would fix this issue

avallete:
> What I was referring to does work (see: https://docs.postgrest.org/en/latest/references/api/computed_fields.html#computed-fields), my question is about fixing the generated type as the typescript error message is preventing builds of my next.js app to succeed. I've been forced to add ts-ignore everywhere since 2 years and I was thinking that this PR sounds like it would fix this issue

Thank's for pointing out the case, I'll have a look, if it's correctly picked up as a column during introspection then it should work. Otherwise, if it's considered a function I don't think the current version will fix this issue. I'll make a test case to make sure of it tough.

whollacsek:
I see in any case thank you for taking the time to improve this part of supabase SDK!

avallete:
I'm locking the discussion to keep the focus on the PR review. Please open an issue or comment over a current one if you want to discuss details.

src/select-query-parser/utils.ts
soedirgo: Reverted this change since as you said this should be handled at the typegen level. This will be fixed in https://github.com/supabase/postgres-meta/pull/809.

> But I think this is actually a much deeper issue about how we handle things, right now everything is focused over a single "schema" so "cross-schema" relations won't work type-wise

PostgREST doesn't detect relationships across schemas, which is why organize the logic within single schemas.

avallete: 👍

soedirgo (approved):

Issue #551, opened 2024-07-11
Unable to import helper types since 1.15.6


olee:
# Bug report

## Describe the bug

Before 1.15.6 it was possible to import helper types from `@supabase/postgrest-js` to perform type mappings etc. like this:
```ts
import type { GetResult } from "@supabase/postgrest-js/dist/module/select-query-parser";

export type TablesSelect<
    Schema extends GenericSchema,
    TableName extends keyof Schema['Tables'],
    Query extends string,
> = GetResult<
    Schema,
    Schema['Tables'][TableName]['Row'],
    TableName,
    Schema['Tables'][TableName] extends { Relationships: infer R } ? R : never,
    Query
>;
```

However, starting from version 1.15.6 with [the change to the package build (package.json etc)](https://github.com/supabase/postgrest-js/commit/2c4e846c8f697f760e15e06a60014e8b93bf3b34#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R19-R22), it became impossible to import any files from subdirectories of the build because of the exports configuration.

## Expected behavior

All types of the library which are publicly exposed as part of generics, function arguments and whatsoever should be properly exported imho so that it is possible to properly type code depending on postgrest-js.

The easiest solution would probably be to export helper types through some namespace from index.ts.

soedirgo:
Hi! These helper types have always been internal and subject to breaking changes - they're not part of the API. Can you share your use case and why you need them?

olee:
The use case is exactly as described above: To generate types for table selects.

For example to write a component that accepts some data which has been selected somewhere else:

```ts
function MyUser(props: { entity: TablesSelect<'users', 'id,first_name,last_name,avatar'> }) {
  return (
    <div>...render user here...</div>
  );
}

// ... at some other location
await supabaseClient.from('users').select('id,first_name,last_name,avatar')....;
```

This is extremly useful and currently prevents us from upgrading supabase client as this is not possible anymore now.

tobico:
I'm also using `GetResult` extensively as the data fetched from my queries is passed around various places in my frontend and I want to correctly type all components and functions that process it.

cco3:
If I understand correctly, there were three way to get result types:

1. Use GetResult, etc.
2. Run the query and get the type of the result
3. Wrap a query in a function and get the return type of the function

1 wasn't supported and was somewhat painful.  2 doesn't give much flexibility for creating helper functions etc. 3 is heavy.

[...]

PR #560, opened 2024-10-12
Re-exports GetResult and Generic helpers to type Postgrest Queries

uncvrd:

Hi all! This PR simply re-exports a couple helper types that many of us were using to strongly type our query responses

Please let me know if I can help improve this!

Ever since 1.15.6, these types had been removed, they were internal but recent discussion is that these should be re-exposed

It used to be

```
import { GenericSchema, GenericTable } from "@supabase/postgrest-js/dist/module/types";
import { GetResult } from "@supabase/postgrest-js/dist/module/select-query-parser";
```

See #551 

## What is the new behavior?

It is now:

```
import { GetResult, GenericSchema, GenericTable } from "@supabase/postgrest-js";
```

In order to do something like the following:

```
type Collection = SelectPublic<"Collection", "id,imageObjectId,type,name,workspaceId">
```

We can build a helper type like this:

```
export type SchemaKeyType = keyof DB;
export type DatabaseTableKeyType = keyof DB["public"]["Tables"];
type RawSelect<
    SchemaName extends SchemaKeyType,
    TableName extends DatabaseTableKeyType,
    Query extends string = "*",
    Schema extends GenericSchema = DB[SchemaName],
    Table extends GenericTable = Schema["Tables"][TableName],
    Relationships = Table extends { Relationships: infer R } ? R : unknown
> = GetResult<Schema, Table["Row"], TableName, Relationships, Query>;

export type SelectPublic<Table extends DatabaseTableKeyType, Query extends string = "*"> = RawSelect<"public", Table, Query>;
```

In order to do this we need access to: `GetResult`, `GenericSchema`, and `GenericTable`

So I've simply exported those from their respective files.

Now we get our get our types back :~)

image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,3 +29,6 @@ export type {
   PostgrestSingleResponse,
   PostgrestMaybeSingleResponse,
 } from './types'
+// https://github.com/supabase/postgrest-js/issues/551
+// To be replaced with a helper type that only uses public types
+export type { GetResult as UnstableGetResult } from './select-query-parser/result'


alexgorbatchev:
@olee @soedirgo please merge this

olee:
@alexgorbatchev afaik the GetResult type should still be considered internal and there was no intention to export it. (see [this comment](https://github.com/supabase/postgrest-js/issues/551#issuecomment-2344986575)).
Instead, a proper helper type should be exported publicly, like eg. this:
```ts
// Exported from postgrest-js
export type TablesSelect<
    Schema extends GenericSchema,
    TableName extends ???,
    Query extends string = '*'
> = ...;

// usage:
type MyEntitySelected = TablesSelect<Database['public'], 'my_table', 'id,title'>;
```

However, as the current state of the library causes quite some issues with various users and even libraries depending on the exported `GetResult` I'd suggest exporting it with an unstable prefix eg. `Unstable_GetResult<...>`.
What do you think @soedirgo?

soedirgo:
Hey all, thanks for the wait. I couldn't come up with a stable helper type in time so I've added an unstable helper type as per @olee's suggestion.

I've omitted `GenericTable` and `GenericSchema` since these are also internal types and easily reimplemented - [here's an example](https://github.com/supabase/supabase-js/blob/456f27e02e70866b2ae5536e3f1f540a55d4a405/src/lib/types.ts#L83).

soedirgo (approved):

uncvrd:
@soedirgo thanks a lot for updating this, I just upgraded : ) fwiw for anyone else reading here, make sure to check the master branch of the example linked above as the Generic types have changed (I got stuck on this)

PR #570, opened 2024-10-31
fix(types): types result with schema as any

avallete:

Fixes the regression in the types inferences if the client is called without a valid schema and defaulted to any

Context: https://discord.com/channels/839993398554656828/1300816893682126912/1301172995326476288

## What is the new behavior?

## Additional context

### DIFF
--- a/src/select-query-parser/result.ts
+++ b/src/select-query-parser/result.ts
@@ -13,6 +13,7 @@ import {
 import {
   CheckDuplicateEmbededReference,
   GetFieldNodeResultName,
+  IsAny,
   IsRelationNullable,
   ResolveRelationship,
   SelectQueryError,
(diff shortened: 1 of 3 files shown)


soedirgo:
I'd rather we keep the previous behavior and just return `any` for the whole result if `Database` is `any` - we never intended type safety to apply if users don't supply database types.

avallete:
> I'd rather we keep the previous behavior and just return `any` for the whole result if `Database` is `any` - we never intended type safety to apply if users don't supply database types.

We totally can. Just to make sure what you say "keep the previous behavior" you mean the "new previous one" ? 😅 

Because what I reproduced here was based on the results were with an `any` Database in [v1.16.2](https://github.com/supabase/postgrest-js/releases/tag/v1.16.2).

So we can decide to change this in favor of "everything is any" but it'll probably be considered as a breaking change by the users so maybe we wanna warn them somehow, or make it clear in the release that this was undesired behavior before.

soedirgo:
I meant the old one before the select-query-parser rewrite - it should be returning the whole result type as `any`:

Screenshot 2024-11-05 at 3 59 49 PM

avallete:
> I meant the old one before the select-query-parser rewrite - it should be returning the whole result type as `any`:
> 
> Screenshot 2024-11-05 at 3 59 49 PM
> (This is v1.16.3)

Yes but not only 😅  everything that's into `test/select-query-parser/default-inference-d.ts` was working the same way in `1.16.3` from my testing, which is why I reproduced the behavior in the first place:

Screenshot 2024-11-05 at 09 20 02
Screenshot 2024-11-05 at 09 20 12

soedirgo:
Ahh ok got it - yeah let's preserve this behavior then 👍

soedirgo (approved):

PR #571, opened 2024-11-01
chore(types): make types retrocompatible to typescript 4.5.5

avallete:

One of the probable reason causing the number of issues we had with the new inference is probably the use in our tests of typescript `~4.7`, in `supabase-js` we can see the version of typescript is `4.5.5`.

This PR is here to reduce the migration friction by allowing the new parser to be retro-compatible with lower typescript version, down to 4.5.5 (I tested down to `4.3.3`).

The main change across all was to get rid of the new convenient `infer X extends Y` syntax only available in typescript 4.7 and higher.

## What is the new behavior?

## Additional context


package.json
```diff
@@ -60,7 +60,7 @@
     "ts-jest": "^28.0.3",
     "tsd": "^0.31.2",
     "typedoc": "^0.22.16",
-    "typescript": "~4.7",
```
avallete, line 63: **remark**

We should set this to the lowest version we want to support in both `supabase-js` and here ?

Or maybe there is another way to run our types tests across multiples major versions of typescript so we can check if there is a breaking change in one of them ? 🤔 

I'm open to suggestions here.

So we reduce the chances of introducing issues in user codebases with latest TS syntax.

soedirgo: I think pinning it to the oldest version we support is sufficient - TS makes breaking changes all the time but afaik we haven't bumped into an issue caused by that

avallete: So we pin it down to `4.5.5` like I did so it matches what we have on `supabase-js` side ? Sounds good ?
Otherwise, what's the lowest version we wanna support ?

soedirgo: Yep, `4.5.5.` should be enough

soedirgo:
@avallete can you limit the change to `infer X extends Y` replacements? There's a bunch of other changes here that seem unrelated (`never` replacements, `Node` -> `Node['type']`, etc.)

avallete:
> @avallete can you limit the change to `infer X extends Y` replacements? There's a bunch of other changes here that seem unrelated (`never` replacements, `Node` -> `Node['type']`, etc.)

I'll try but:

1. The `Node -> Node['type']` change is here because the PR is based over [this one](https://github.com/supabase/postgrest-js/pull/569) where I responded [here](https://github.com/supabase/postgrest-js/pull/569#discussion_r1822333370) about why I think it's still necessary.

2. I'll look into the never I mainly did it to ease the reporting and debugging process (it's easier to look for a single unique error string in a specific line than "any never" in the code), and only kept it at place where changing it would have required logic changes somewhere else. But I can revert that and keep the `never` if you confirm that's what you want those are the places where I have introduced "unrelated never":

https://github.com/supabase/postgrest-js/pull/571/files#diff-5e58fb7db093dc508e6abb63dd5488067ce801a1abbcb5a50644050d106ff1edL174-R179
https://github.com/supabase/postgrest-js/pull/571/files#diff-5e58fb7db093dc508e6abb63dd5488067ce801a1abbcb5a50644050d106ff1edL259-R266

3. Also note that some of the changes notably in the tests is because of downgrading TS to 4.5.5, where `function() !== function(undefined)` and null narrowing wasn't the same.

avallete:
@soedirgo 

I've rebased this one over the correct branch so the diff should be much more focus on the actual changes for 4.5.5 compatibility.

Also reverted some noise with the never replacement / new errors that wasn't necessary to the conversion.

Note that this PR should be merged after the one it base itself of: https://github.com/supabase/postgrest-js/pull/569

soedirgo (approved):

PR #577, opened 2024-11-13
fix(types): inference for default any schema

avallete:

- Detected a bug for the "any" schema case while upgrading the workers dependency this make the inference logic more recursive reducing the chances of error and restore the type to match what was generated on `2.46.1`

## What is the new behavior?

## Additional context

### DIFF
--- a/src/select-query-parser/result.ts
+++ b/src/select-query-parser/result.ts
@@ -78,15 +78,9 @@ type ProcessFieldNodeWithoutSchema<Node extends Ast.FieldNode> = IsNonEmptyArray
   Node['children']
 > extends true
   ? {
-      [K in Node['name']]: Node['children'] extends Ast.StarNode[]
-        ? any[]
-        : Node['children'] extends Ast.FieldNode[]
-        ? {
-            [P in Node['children'][number] as GetFieldNodeResultName<P>]: P['castType'] extends PostgreSQLTypes
-              ? TypeScriptTypes<P['castType']>
-              : any
-          }[]
-        : any[]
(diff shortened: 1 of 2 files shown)


src/select-query-parser/result.ts
```diff
@@ -78,15 +78,9 @@ type ProcessFieldNodeWithoutSchema<Node extends Ast.FieldNode> = IsNonEmptyArray
...
-      [K in Node['name']]: Node['children'] extends Ast.StarNode[]
-        ? any[]
-        : Node['children'] extends Ast.FieldNode[]
-        ? {
-            [P in Node['children'][number] as GetFieldNodeResultName<P>]: P['castType'] extends PostgreSQLTypes
-              ? TypeScriptTypes<P['castType']>
-              : any
-          }[]
-        : any[]
+      [K in GetFieldNodeResultName<Node>]: Node['children'] extends Ast.Node[]
```
avallete, line 81: **remark**

Use `GetFieldNodeResultName` to have the embedding properly renamed.

soedirgo (approved):

src/select-query-parser/result.ts
```diff
@@ -78,15 +78,9 @@ type ProcessFieldNodeWithoutSchema<Node extends Ast.FieldNode> = IsNonEmptyArray
...
+        ? ProcessNodesWithoutSchema<Node['children']>[]
+        : ProcessSimpleFieldWithoutSchema<Node>
```
avallete, line 83: **remark**

Re-use the other functions to infer the final result recursively rather than duplicating the logic.

PR #578, opened 2024-11-17
fix: Add missing PostgrestError export to wrapper.mjs

krizzu:

As per title

`PostgrestError` is not properly exported

## What is the new behavior?

`PostgrestError` is properly exported


soedirgo (approved):
Thanks!

PR #579, opened 2024-11-23
chore: rename type test file

mrazauskas:

Only a file rename.

The `.test/select-query-parser/default-inference-d.ts` file is not checked by `tsd`, because its name does not match the specified pattern:

https://github.com/supabase/postgrest-js/blob/0af65f55da930dc0b13d18a5c6994a72540c79f0/package.json#L43

You can proof this by adding a failing assertion to the file and running `npm run test:types`. The error is not reported.

## What is the new behavior?

Renaming the file to `./test/select-query-parser/default-inference.test-d.ts` makes sure it is picked up by `tsd`.

## Additional context

Unfortunately `tsd` does not print names of test files. That makes it very easy to make similar mistakes.


soedirgo (approved):
Missed that, thanks!

Issue #351, opened 2022-10-14
NO Typing/ autocompletion hint working When filtering on Foreign keys -  V2. 


roker15:
Here is [Example link](https://supabase.com/docs/reference/javascript/select#:~:text=table%20is%20not-,filtered,-out.) from supabase docs. 

```
const { data, error } = await supabase
  .from('cities')
  .select('name, countries(*)')
  .eq('countries.name', 'Estonia')
```
In above example **`countries.name`** shows error in types.  And no autocompletion available after` countries.`

avallete:
I've been able to reproduce similar with tests on the repo like so:

```ts
test('select all channels details', async () => {
  const res = await postgrest.from('channel_details').select('id, channels(*)')
  expect(res).toMatchInlineSnapshot(`
    Object {
      "count": null,
      "data": Array [
        Object {
          "channels": Object {
            "data": null,
            "id": 1,
            "slug": "public",
          },
          "id": 1,
        },
        Object {
          "channels": Object {
            "data": null,
            "id": 2,
            "slug": "random",
          },
          "id": 2,
        },
      ],
      "error": null,
      "status": 200,
      "statusText": "OK",
    }
  `)
})

test('select filter by id', async () => {
  const res = await postgrest.from('channel_details').select('id, channels(*)').eq('id', 1)
  expect(res).toMatchInlineSnapshot(`
    Object {
      "count": null,
      "data": Array [
        Object {
          "channels": Object {
            "data": null,
            "id": 1,
            "slug": "public",
          },
          "id": 1,
        },
      ],
      "error": null,
      "status": 200,
      "statusText": "OK",
    }
  `)
})

test('select filter by channels.id', async () => {
  const res = await postgrest.from('channel_details').select('id, channels(*)').eq('channels.id', 1)
  expect(res).toMatchInlineSnapshot(`
    Object {
      "count": null,
      "data": Array [
        Object {
          "channels": Object {
            "data": null,
            "id": 1,
            "slug": "public",
          },
          "id": 1,
        },
        Object {
          "channels": null,
          "id": 2,
        },
      ],
      "error": null,
      "status": 200,
      "statusText": "OK",
    }
  `)
})
```

But I actually think that the behavior of postgrest is very weird to follow here. I would expect the last case, to give me the same result as the 1st one. But instead, postgrest apply the condition over the children only and pull null into the embedded relation that doesn't match the filter. Which by the way break the types since channels is actually a non-nullable.

Issue #169, opened 2021-03-29
Typing is incorrect when filtering by embedded attributes


samtgarson:
# Bug report

## Describe the bug

Postgrest allows [filtering by attributes on embedded resources](https://postgrest.org/en/v7.0.0/api.html#embedded-filters), but the typings on the filter functions only support attributes on the selected table.

## To Reproduce

This [recreation](https://codesandbox.io/s/hungry-tdd-vy3xb?file=/src/index.ts) would work when run with valid credentials, but the typings fail and cause a build error.

## Expected behavior

Ideally, the function signatures would support type-safe paths in dot-notation, but it should at least not cause a build error.

[...]

PR #589, opened 2024-12-31
fix: Correctly validate relationship enum values in eq, neq and in methods

kamilogorek:
Builds on top of https://github.com/supabase/postgrest-js/pull/588/

I couldn't find a nice way to remove duplication from `ResolveFilterValue` calls, and we need to branch out there, as `ResolveFilterValue` is sometimes wrapped in `NonNullable`, sometimes in `ReadonlyArray` and sometimes in nothing. Fallback values are also different.

In perfect world this would work, but I don't know hot to express a type for `Wrapper` as a generic accepting a single type 🤔 

```ts
type ResolveFilterValue<
  Tables extends Record<string, GenericTable>,
  Row extends Record<string, unknown>,
  ColumnName extends string,
  Wrapper,
  Fallback
> = ColumnName extends `${infer RelationshipTable}.${infer RelationshipColumn}`
  ? ResolveFilterRelationshipValue<Tables, RelationshipTable, RelationshipColumn>
  : ColumnName extends keyof Row
  ? Wrapper extends Something
    ? Wrapper<Row[ColumnName]>
    : Row[ColumnName]
  : Fallback
```

Fixes https://github.com/supabase/postgrest-js/issues/169
Fixes https://github.com/supabase/postgrest-js/issues/351

I guess?™️


test/index.test-d.ts
```diff
@@ -21,12 +21,26 @@ const postgrest = new PostgrestClient<Database>(REST_URL)
...
+// `.eq()`, '.neq()' and `.in()` validate provided filter value when column is an enum.
+// Behaves the same for simple columns, as well as relationship filters.
 {
   expectError(postgrest.from('users').select().eq('status', 'invalid'))
   expectError(postgrest.from('users').select().neq('status', 'invalid'))
   expectError(postgrest.from('users').select().in('status', ['invalid']))
 
+  expectError(
+    postgrest.from('best_friends').select('users!first_user(status)').eq('users.status', 'invalid')
+  )
```
avallete, line 33: **question**

Might this break some current codebase ? Where the filter isn't declared as a plain string, but it rather a variable, if the variable value isn't properly narrowed down by typescript.

So we might want to keep an eye out for this kind of errors / mention it in the release note.

src/PostgrestFilterBuilder.ts
```diff
@@ -25,6 +25,30 @@ type FilterOperator =
   | 'phfts'
   | 'wfts'
 
+// Match relationshop filters with `table.column` syntax and resolve underlying column value.
```
avallete, line 28: **nitpick**

```suggestion
// Match relationship filters with `table.column` syntax and resolve underlying column value.
```

test/index.test-d.ts
kamilogorek: Filter value or the key? Value will always fallback to whatever `extends never` branch evaluates to, so either `unknown` or `unknown[]`. As for key, it was always typed as `string` in all previous releases.

soedirgo: Yep, so if the `status` variable is a `string` but the column is an enum, it will now error out.

+1 on just carrying on and mentioning it in the release notes, I think it's still a desirable change to make it stricter as long as there's no false negatives.

kamilogorek: Ah gotcha. Then yes, I think it can break under some conditions, although I'm not exactly sure if it should be treated as "breaking" or rather as a "bug fix".
If you think about it, it'll now error when someone passes a number but type expected string and other way around. The same goes for someone passing `"some-value"` to explicit string union type.
The only "valid" usecase which it can actually break is exactly the one you provided.
However, if someone is using generated types (which is the only people this change affects), then this function should be typed as:

```ts
function getAllStatuses(status: Enums<'statuses'>) {
 postgrest.from('table').select('*').eq('status', status)
}
```

Soooo breaking? yes. Expected? Probably also yes. I don't see a way to "not break", as it'd require relaxing types, which is a whole purpose of this change.

src/PostgrestFilterBuilder.ts
```diff
@@ -25,6 +25,30 @@ type FilterOperator =
   | 'phfts'
   | 'wfts'
 
+// Match relationship filters with `table.column` syntax and resolve underlying column value.
+// If not matched, fallback to generic type.
+type ResolveFilterValue<
```
soedirgo, line 30: Might have to make this recursive to account for deeply nested embedded tables, e.g. `.eq('messages.channels.id', 42)`

test/index.test-d.ts
avallete: Totally agree @kamilogorek but I learned the hard way that people are not happy when you "fix" unexpected behavior they rely on in their codebase 😂 

Especially since postgrest would not error in such case, just probably return an empty result, so that make runtime / type-check behavior a bit weird to grasp for the users sometimes.

I think mentioning it in the release note (so community/support can link to it) and adding an example of how to "fix the issue" with a cast or something like @soedirgo mentioned will be more than enough for this. It'll just preshot some of the "issues" we might see being opened after users upgrade.

src/PostgrestFilterBuilder.ts
soedirgo: Resolved in https://github.com/supabase/postgrest-js/pull/589/commits/697cc601d3eb31496b95e86bd705d5e6e861558b

test/index.test-d.ts
kamilogorek: Oh definitely, I supported Sentry JS SDK for way too many years to learn that the hard way.
Although tbh if someone opts-in to the type checking, I'd say they _should_ be happy that it became better 🤷🏻‍♂️

src/PostgrestFilterBuilder.ts
kamilogorek: <3

test/index.test-d.ts
kamilogorek: I can dogfood it on our own infra repo before we merge the change into supabase-js as well.

avallete (approved):

PR #592, opened 2025-01-12
feat(types): add json path type inference

avallete:

Allow users to override Json type with custom types, and take into account the "json accessor" to infer the proper resulting type.

Given a table defined like so:

```
type CustomJsonType = {
  foo: string
  bar: {
    baz: number
  }
  en: 'ONE' | 'TWO' | 'THREE'
}
table: {
 data: CustomJsonType | null
}
```

The following behavior is observed:

```ts
select(`data`) -> {data: CustomJsonType | null}
select(`data->bar`) -> {bar: Json}
select(`data->>bar`) -> {bar: string}
select(`data->>en`) -> {bar: string}
select(`data->bar->baz`) -> {baz: Json}
```

## What is the new behavior?

```ts
select(`data`) -> {data: CustomJsonType | null}
select(`data->bar`) -> { bar:  { baz: number } }
// Preserve the postgrest casting behaviour due to ->> accessor
select(`data->>bar`) -> {bar: string}
// Preserve the union of string
select(`data->>en`) -> {bar:  'ONE' | 'TWO' | 'THREE'}
// Preserve the accessed ressource type
select(`data->bar->baz`) -> {baz: number}
// Fallback on the default behavior for non-matching properties
select(`data->bar->nope`) -> {nope: Json}
select(`data->bar->>nope`) -> {nope: string}
```

## Additional context

### DIFF
--- a/src/select-query-parser/parser.ts
+++ b/src/select-query-parser/parser.ts
@@ -2,6 +2,7 @@
 // See https://github.com/PostgREST/postgrest/blob/2f91853cb1de18944a4556df09e52450b881cfb3/src/PostgREST/ApiRequest/QueryParams.hs#L282-L284
 
 import { SimplifyDeep } from '../types'
+import { JsonPathToAccessor } from './utils'
 
 /**
  * Parses a query.
@@ -220,13 +221,24 @@ type ParseNonEmbeddedResourceField<Input extends string> = ParseIdentifier<Input
 ]
   ? // Parse optional JSON path.
     (
-      Remainder extends `->${infer _}`
+      Remainder extends `->${infer PathAndRest}`
         ? ParseJsonAccessor<Remainder> extends [
             infer PropertyName,
             infer PropertyType,
             `${infer Remainder}`
           ]
-          ? [{ type: 'field'; name: Name; alias: PropertyName; castType: PropertyType }, Remainder]
+          ? [
+              {
+                type: 'field'
+                name: Name
+                alias: PropertyName
+                castType: PropertyType
+                jsonPath: JsonPathToAccessor<
+                  PathAndRest extends `${infer Path},${string}` ? Path : PathAndRest
+                >
+              },
+              Remainder
+            ]
           : ParseJsonAccessor<Remainder>
         : [{ type: 'field'; name: Name }, Remainder]
     ) extends infer Parsed
@@ -401,6 +413,7 @@ export namespace Ast {
     hint?: string
     innerJoin?: true
     castType?: string
+    jsonPath?: string
     aggregateFunction?: Token.AggregateFunction
     children?: Node[]
   }
--- a/src/select-query-parser/utils.ts
+++ b/src/select-query-parser/utils.ts
@@ -546,9 +546,33 @@ export type FindFieldMatchingRelationships<
   : SelectQueryError<'Failed to find matching relation via name'>
 
 export type JsonPathToAccessor<Path extends string> = Path extends `${infer P1}->${infer P2}`
-  ? P2 extends `>${infer Rest}` // Check if P2 starts with > (from ->>)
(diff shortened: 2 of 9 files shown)


src/select-query-parser/parser.ts
```diff
@@ -220,13 +221,24 @@ type ParseNonEmbeddedResourceField<Input extends string> = ParseIdentifier<Input
...
             `${infer Remainder}`
           ]
-          ? [{ type: 'field'; name: Name; alias: PropertyName; castType: PropertyType }, Remainder]
+          ? [
+              {
+                type: 'field'
+                name: Name
+                alias: PropertyName
+                castType: PropertyType
+                jsonPath: JsonPathToAccessor<
```
avallete, line 236: We use this to format and save the value "path accessor" information from the parser to the result inference. That way, the parser can still be agnostic to the database types, only caring about string parsing.

src/select-query-parser/utils.ts
```diff
@@ -544,3 +544,35 @@ export type FindFieldMatchingRelationships<
       name: Field['name']
     }
   : SelectQueryError<'Failed to find matching relation via name'>
+
+export type JsonPathToAccessor<Path extends string> = Path extends `${infer P1}->${infer P2}`
```
avallete, line 548: Turn something like `data->some->>property` into `some.property`

src/select-query-parser/utils.ts
```diff
@@ -544,3 +544,35 @@ export type FindFieldMatchingRelationships<
...
+    : P2 extends string // Handle -> operator
+    ? JsonPathToAccessor<`${P1}.${P2}`>
+    : Path
+  : Path extends `>${infer Rest}` // Clean up any remaining > characters
+  ? JsonPathToAccessor<Rest>
+  : Path extends `${infer P1}::${infer _}` // Handle type casting
+  ? JsonPathToAccessor<P1>
+  : Path
+
+export type JsonPathToType<T, Path extends string> = Path extends ''
```
avallete, line 560: Parse some path like `some.property` and try to match it with the type T provided.

test/index.test-d.ts
```diff
@@ -52,87 +52,93 @@ const postgrest = new PostgrestClient<Database>(REST_URL)
   )
 
   {
-    const { data, error } = await postgrest.from('users').select('status').eq('status', 'ONLINE')
-    if (error) {
-      throw new Error(error.message)
+    const result = await postgrest.from('users').select('status').eq('status', 'ONLINE')
```
avallete, line 55: Here for some reason (maybe TS upgrade) using this syntax:

```
const { data, error } = await ...

if (error) {
 throw XXX
}
// here data was still of type data[] | null
// because typescript seemed to fail to understand the branching
```

Changing to this syntax properly narrow down `result.data` as non nullable once we checked the error.

kamilogorek:
That's already a black-magic level, but validated behavior on our live project and everything works as expected.

kamilogorek:
Gentle ping @soedirgo

soedirgo:
Seems to not work on embedded tables:

```
    const result = await postgrest
      .from('messages')
      .select('users(data->bar)')
```

(I updated the types for public.users.data with `CustomUserDataType`)

But that can be handled in a separate PR 👍

soedirgo (approved):

avallete:
> Seems to not work on embedded tables:
> 
> ```
>     const result = await postgrest
>       .from('messages')
>       .select('users(data->bar)')
> ```
> 
> (I updated the types for public.users.data with `CustomUserDataType`)
> 
> But that can be handled in a separate PR 👍

Good catch, it was an issue with the parser and the closing parenthesis, I fixed it here: https://github.com/supabase/postgrest-js/pull/592/commits/a9842b27240948a1dbe3fb602af597f13584a951

And added appropriate tests for coverage 👍

PR #593, opened 2025-01-13
types: export GenericSchema type from types

brunorodmoreira:

Add the export of `GenericSchema` in order to allow to use it as base to Database type argument in the PostgrestClient.

The type is not exported, so if we are using the client without the generated types, we are not able to get type hints.

## What is the new behavior?

Exporting it, we can extend it and get type hints.

## Additional context

Expected behavior: image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,6 +28,7 @@ export type {
   PostgrestResponseSuccess,
   PostgrestSingleResponse,
   PostgrestMaybeSingleResponse,
+  GenericSchema,
 } from './types'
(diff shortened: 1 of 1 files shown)


mandarini:
Hi @brunorodmoreira! Thank you very much for the time and effort you put in this contribution. I tried to import your PR in the new repo (https://github.com/supabase/supabase-js) but there were merge conflicts, since the code has changed a lot in the past year. Could you please take the time to create a new PR in the new repo (https://github.com/supabase/supabase-js), if you still think this change is valuable?

Thank you once again for contributing to Supabase! 💚

PR #597, opened 2025-01-20
fix(types): type instantiation is excessively deep and possibly infinite

avallete:

Bug fix a case on typescript > 4.5.5 where using `MergeDeep` override lead to an error of `Type instantiation is excessively deep and possibly infinite`. Use a trick to "force typescript to infer final type" instead of doing a shallow inspection and warning about possible infinite recursion.

This is a bit black magic'ish but it fix the issue on the reproduction case 😅 

## Additional context

Related: https://github.com/supabase/supabase-js/issues/1354

### DIFF
--- a/src/PostgrestFilterBuilder.ts
+++ b/src/PostgrestFilterBuilder.ts
@@ -273,11 +273,13 @@ export default class PostgrestFilterBuilder<
    */
   in<ColumnName extends string>(
     column: ColumnName,
-    values: ResolveFilterValue<Schema, Row, ColumnName> extends infer ResolvedFilterValue
-      ? ResolvedFilterValue extends never
-        ? unknown[]
-        : ReadonlyArray<ResolvedFilterValue>
-      : never
+    values: ReadonlyArray<
+      ResolveFilterValue<Schema, Row, ColumnName> extends infer ResolvedFilterValue
+        ? ResolvedFilterValue extends never
+          ? unknown[]
+          : ResolvedFilterValue
+        : never
+    >
   ): this {
     const cleanedValues = Array.from(new Set(values))
       .map((s) => {
--- a/src/PostgrestFilterBuilder.ts
+++ b/src/PostgrestFilterBuilder.ts
@@ -75,9 +75,9 @@ export default class PostgrestFilterBuilder<
     column: ColumnName,
     value: ResolveFilterValue<Schema, Row, ColumnName> extends infer ResolvedFilterValue
       ? ResolvedFilterValue extends never
-        ? NonNullable<unknown>
+        ? unknown
         : NonNullable<ResolvedFilterValue>
-      : never
+      : unknown
   ): this {
     this.url.searchParams.append(column, `eq.${value}`)
     return this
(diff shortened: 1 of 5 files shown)


src/PostgrestFilterBuilder.ts
```diff
@@ -73,9 +73,11 @@ export default class PostgrestFilterBuilder<
...
   eq<ColumnName extends string>(
     column: ColumnName,
-    value: ResolveFilterValue<Schema, Row, ColumnName> extends never
-      ? NonNullable<unknown>
-      : NonNullable<ResolveFilterValue<Schema, Row, ColumnName>>
+    value: ResolveFilterValue<Schema, Row, ColumnName> extends infer ResolvedFilterValue
+      ? ResolvedFilterValue extends never
+        ? NonNullable<unknown>
+        : NonNullable<ResolvedFilterValue>
+      : never
```
avallete, line 80: The actual tricks is to use the `infer` and `never` to force typescript to infer the right type and not stop at "possibly infinite recursive".

src/PostgrestFilterBuilder.ts
```diff
@@ -269,9 +273,13 @@ export default class PostgrestFilterBuilder<
...
-    values: ResolveFilterValue<Schema, Row, ColumnName> extends never
-      ? unknown[]
-      : ReadonlyArray<ResolveFilterValue<Schema, Row, ColumnName>>
+    values: ReadonlyArray<
+      ResolveFilterValue<Schema, Row, ColumnName> extends infer ResolvedFilterValue
+        ? ResolvedFilterValue extends never
+          ? unknown[]
+          : ResolvedFilterValue
+        : never
+    >
```
avallete, line 282: Here we need the `infer` within the `ReadonlyArray` to have the proper resulting type, otherwise we'll have an invalid union.

test/relationships.ts
```diff
@@ -1830,7 +1830,12 @@ test('self reference relation via column', async () => {
 })
 
 test('aggregate on missing column with alias', async () => {
-  const res = await selectQueries.aggregateOnMissingColumnWithAlias.eq('id', 1).limit(1).single()
+  const res = await selectQueries.aggregateOnMissingColumnWithAlias
+    // @ts-expect-error should not be able to eq 'id' since the column does not
+    // exist
+    .eq('id', 2)
```
avallete, line 1836: I think this test should never had passed because the query actually is made to return an error. And `id` column doesn't exist in `users` table.

We can make it pass tough by changing the `never` into an `unknown` here:

https://github.com/supabase/postgrest-js/pull/597/files#diff-8c09e973443260962548f07f6714d574185bce9a909a6d818691993e5a4253e5L43

Eigilak:
Not All Heroes Wear Capes 🫡

avallete:
Tested a build of `supabase-js` using this new filter types over our internal api codebase. Worked without raising types errors and didn't seems to impact the inference types much (`--extendedDiagnostics` raised very similar results).

kamilogorek (approved):
image

PR #603, opened 2025-02-17
feat(types): add override types method

avallete:
- Add `overrideTypes` method allow both partial and complete override of the return type, do partial override by default

@kamilogorek mentioned that we often had to port a lot of properties definition in order to override a single field when using `.returns` method.

This new method introduce a way to perform "partial override" per default, and allow the user to have similar behavior than the current `.returns` method by providing the option `{ merge: false }`.

This make this kind of type override where we only want to override `func_call` result easier to maintain.

**Before**:
```ts
const results = await getOrThrowTyped(
  readWriteNext
    .from('sometable')
    .select(
      'id, projects!projects_sometable_id_fkey(ref, name), func_call(id)'
    )
    .eq('slug', 'some-slug')
    .is('projects.metadata->base_project_id', null)
    .neq('projects.status', 'SOMEVALUE')
    .single()
    .returns<{
      id: number
      func_call?: { value: SomeEnum }
      projects: { ref: string; name: string }[]
    }>(),
  'Failed to retrieve infos'
)
```

**After**

```ts
const results = await getOrThrowTyped(
  readWriteNext
    .from('sometable')
    .select(
      'id, projects!projects_sometable_id_fkey(ref, name), func_call(id)'
    )
    .eq('slug', 'some-slug')
    .is('projects.metadata->base_project_id', null)
    .neq('projects.status', 'SOMEVALUE')
    .single()
    .overrideTypes<{
      // Here we can now override only the part that we are interested in
      func_call?: { value: SomeEnum }
    }>(),
  'Failed to retrieve infos'
)

// We can still opt out from the merging behavior with:
await getOrThrowTyped(
  readWriteNext
    .from('sometable')
    .select(
      'id, projects!projects_sometable_id_fkey(ref, name), func_call(id)'
    )
    .eq('slug', 'some-slug')
    .is('projects.metadata->base_project_id', null)
    .neq('projects.status', 'SOMEVALUE')
    .single()
    .overrideTypes<{
      id: number
      func_call?: { value: SomeEnum }
      projects: { ref: string; name: string }[]
    }, { merge: false }>(),
  'Failed to retrieve infos'
)
```

### DIFF
--- a/src/PostgrestBuilder.ts
+++ b/src/PostgrestBuilder.ts
@@ -6,6 +6,8 @@ import type {
   PostgrestSingleResponse,
   PostgrestResponseSuccess,
   CheckMatchingArrayTypes,
+  MergePartialResult,
+  IsValidResultOverride,
 } from './types'
 import PostgrestError from './PostgrestError'
 
@@ -224,6 +226,45 @@ export default abstract class PostgrestBuilder<Result, ThrowOnError extends bool
   returns<NewResult>(): PostgrestBuilder<CheckMatchingArrayTypes<Result, NewResult>, ThrowOnError> {
     return this as unknown as PostgrestBuilder<
       CheckMatchingArrayTypes<Result, NewResult>,
+      ThrowOnError
+    >
+  }
+
+  /**
+   * Override the type of the returned `data` field in the response.
+   *
+   * @typeParam NewResult - The new type to cast the response data to
+   * @typeParam Options - Optional type configuration (defaults to { merge: true })
+   * @typeParam Options.merge - When true, merges the new type with existing return type. When false, replaces the existing types entirely (defaults to true)
+   * @example
+   * ```typescript
+   * // Merge with existing types (default behavior)
+   * const query = supabase
+   *   .from('users')
+   *   .select()
+   *   .overrideTypes<{ custom_field: string }>()
+   *
+   * // Replace existing types completely
+   * const replaceQuery = supabase
+   *   .from('users')
+   *   .select()
+   *   .overrideTypes<{ id: number; name: string }, { merge: false }>()
+   * ```
+   * @returns A PostgrestBuilder instance with the new type
+   */
+  overrideTypes<
+    NewResult,
+    Options extends { merge?: boolean } = { merge: true }
+  >(): PostgrestBuilder<
+    IsValidResultOverride<Result, NewResult, true, false, false> extends true
+      ? MergePartialResult<NewResult, Result, Options>
+      : CheckMatchingArrayTypes<Result, NewResult>,
+    ThrowOnError
+  > {
+    return this as unknown as PostgrestBuilder<
+      IsValidResultOverride<Result, NewResult, true, false, false> extends true
+        ? MergePartialResult<NewResult, Result, Options>
+        : CheckMatchingArrayTypes<Result, NewResult>,
       ThrowOnError
     >
   }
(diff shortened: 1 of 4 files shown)


src/PostgrestBuilder.ts
```diff
@@ -219,12 +221,51 @@ export default abstract class PostgrestBuilder<Result, ThrowOnError extends bool
    * Override the type of the returned `data`.
    *
    * @typeParam NewResult - The new result type to override with
-   * @deprecated Use overrideTypes<yourType, { partial: false }>() method at the end of your call chain instead
+   * @deprecated Use overrideTypes<yourType, { merge: false }>() method at the end of your call chain instead
```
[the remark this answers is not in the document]
avallete: I did not based this branch onto master 😅 

Idea will be to merge both together before merging to main. Just split them up for review file diff convenience.

test/override-types.test-d.ts
```diff
@@ -0,0 +1,125 @@
...
+  const maybeSingleResult = await postgrest
+    .from('users')
+    .select()
+    .maybeSingle()
+    .overrideTypes<{ custom_field: string }>()
+  if (maybeSingleResult.error) {
+    throw new Error(maybeSingleResult.error.message)
+  }
+  let maybeSingleResultType: typeof maybeSingleResult.data
+  expectType<TypeEqual<(typeof maybeSingleResultType)['custom_field'], string>>(true)
```
kamilogorek, line 74: I think this should preserve optionality of the result when overriding.

test/override-types.test-d.ts
```diff
@@ -0,0 +1,125 @@
...
+    .overrideTypes<{ custom_field: string }>()
+  if (singleResult.error) {
+    throw new Error(singleResult.error.message)
+  }
+  let result: typeof singleResult.data
+  expectType<
+    TypeEqual<
+      typeof result,
+      {
+        Error: 'Type mismatch: Cannot cast array result to a single object. Use .returns<Array<YourType>> for array results or .single() to convert the result to a single object'
```
kamilogorek, line 23: Should this point to `overrideTypes` instead of `returns`?

kamilogorek:
Concept looks good to me and I believe it'd fix most our type problems.
Just a general note that https://supabase.com/docs/reference/javascript/typescript-support and https://supabase.com/docs/reference/javascript/db-returns will need to be updated/expanded.

test/override-types.test-d.ts
avallete: Good catch

test/override-types.test-d.ts
avallete: Yeah I tough about this as well.

I was worried about not being able to discriminate between an user opt-in optionality with `maybeSingle` and a value that can just be null and that the user might want to override.

But looking into this further, I guess the top level is always either an array of objects, a non-null object (`single`), or a nullable object (`maybeSingle`). If there is no other cases where the result can be null than the `maybeSingle` we can preserve optionality 🤔

avallete:
> Concept looks good to me and I believe it'd fix most our type problems. Just a general note that https://supabase.com/docs/reference/javascript/typescript-support and https://supabase.com/docs/reference/javascript/db-returns will need to be updated/expanded.

Started to write some draft for the docs here: https://github.com/supabase/supabase/pull/33705

avallete:
> Is this meant to be in draft still?

Nope 😓

soedirgo (approved):

PR #604, opened 2025-02-17
fix(types): returns type casting

avallete:
- Allow to use returns at the end of the call chain after a .single()
- Add deprecation warning to move toward explicit overrideTypes method
- Add cast checking basic logic, an array should be declared if the result is an array, an object if the result is an object

Goes along: 

https://github.com/supabase/postgrest-js/pull/603

### DIFF
--- a/src/PostgrestTransformBuilder.ts
+++ b/src/PostgrestTransformBuilder.ts
@@ -307,7 +307,7 @@ export default class PostgrestTransformBuilder<
    * Override the type of the returned `data`.
    *
    * @typeParam NewResult - The new result type to override with
-   * @deprecated Use overrideTypes<yourType, { partial: false }>() method at the end of your call chain instead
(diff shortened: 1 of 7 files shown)


src/PostgrestTransformBuilder.ts
```diff
@@ -307,18 +307,19 @@ export default class PostgrestTransformBuilder<
    * Override the type of the returned `data`.
    *
    * @typeParam NewResult - The new result type to override with
+   * @deprecated Use overrideTypes<yourType, { partial: false }>() method at the end of your call chain instead
```
soedirgo, line 310:
```suggestion
   * @deprecated Use overrideTypes<yourType, { merge: false }>() method at the end of your call chain instead
```

soedirgo (approved):

soedirgo (approved):

PR #606, opened 2025-02-21
fix(types): overrideTypes work on invalid embeded relation

avallete:
- When a relation doesn't exist, or that type inference fail to resolve it the new type should be the source of truth for the result. In such case, we don't "merge" the properties together, but just replace the error with the user provided type.
- When chaining the override with `.maybeSingle` preserve the optionality of the result
- Allow merging between literal object and non literal one eg:
```ts
// Given some property like
{
   someProperty: Record<string, Json | null>
}
// calling
overrideType<{ someProperty: { literalProperty: string }  }>()
// result in:
{
    [key: string]: Json | null
    someProperty: { literalProperty: string }
} 
```

### DIFF
--- a/src/PostgrestBuilder.ts
+++ b/src/PostgrestBuilder.ts
@@ -10,6 +10,7 @@ import type {
   IsValidResultOverride,
 } from './types'
 import PostgrestError from './PostgrestError'
+import { ContainsNull } from './select-query-parser/types'
 
 export default abstract class PostgrestBuilder<Result, ThrowOnError extends boolean = false>
   implements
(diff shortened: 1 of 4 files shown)


soedirgo (approved):

jdgamble555:
@avallete - Why not have this work with `single()` as well?

J

avallete:
> @avallete - Why not have this work with `single()` as well?
> 
> J

This should work with `single()` as well (test is here: https://github.com/supabase/postgrest-js/blob/1f93ee55871ad86613a33b97562754387b8c0328/test/override-types.test-d.ts#L205-L228)

Difference here is that `maybeSingle` should always result in an optional (`object | null`) while `single()` should give a non null result.

PR #608, opened 2025-03-08
fix(types): outdated error message

rafalnawojczyk:

Slightly changes the output of the error, as when using:
`.overrideTypes<MyType>()`
for an array of results - I was getting error pointing me to use `.returns<Array<YourType>>` which is currently deprecated. 

I've made a message clear for users that gets this error from `.overrideTypes` as well as for `.returns` enjoyers. 

It returns an error message that can be misleading and pointing to use deprecated features.

## Additional context

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -114,7 +114,7 @@ export type CheckMatchingArrayTypes<Result, NewResult> =
         Result,
         NewResult,
         {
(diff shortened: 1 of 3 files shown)


soedirgo (approved):

soedirgo:
Thanks!

PR #610, opened 2025-03-13
Fix: Allow in() filter to accept arrays directly without manual formatting

geoguide:
## **What kind of change does this PR introduce?**
Bug fix

---

## **What is the current behavior?**
Currently, the `in()` filter requires manually formatting arrays before passing them to the query. Instead of accepting a plain array, users must wrap values in parentheses manually, like this:

```ts
query.filter('vehicle.division_id', 'in', `(${filters.divisionIds.join(',')})`)
```

This is inconvenient and inconsistent with how other filters work.

---

## **What is the new behavior?**
This PR updates the `in()` filter method to **accept arrays directly** without requiring manual formatting. It:
- Ensures `values` is a **non-empty array**.
- **Deduplicates values** using `Set`.
- **Automatically wraps reserved characters** (`, ( )`) in quotes if necessary.
- **Removes the need for manually formatting the array** before passing it.

Now, users can write:
```ts
query.filter('trailer.division_id', 'in', filters.divisionIds)
```
And it will correctly format the query as:
```
trailer.division_id=in.(value1,value2,value3)
```

✅ No more manual wrapping!

---

## **Additional context**
This change improves **developer experience** by ensuring consistency across filter methods. It also helps prevent common mistakes when manually formatting values.

Let me know if any additional changes are needed! 🚀

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage
 .nyc_output
 
 .idea
+.DS_Store
--- a/src/PostgrestFilterBuilder.ts
+++ b/src/PostgrestFilterBuilder.ts
@@ -292,22 +292,19 @@ export default class PostgrestFilterBuilder<
     values: ReadonlyArray<
       ResolveFilterValue<Schema, Row, ColumnName> extends never
         ? unknown
-        : // We want to infer the type before wrapping it into a `NonNullable` to avoid too deep
-        // type resolution error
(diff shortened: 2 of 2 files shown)


soedirgo:
Thanks for the PR @geoguide! You can avoid writing the parentheses by using the [`.in()` method](https://supabase.com/docs/reference/javascript/in). As mentioned [here](https://supabase.com/docs/reference/javascript/filter), `.filter()` is an escape hatch and is not meant to have automatic formatting.

PR #612, opened 2025-03-28
fix(types): inner join type on nullable relationships

avallete:
Thank's to MRE example repo setup here: https://github.com/masda70/supabase-non-null-inner-join/tree/working

Fixes the type for `!inner` join that are typed as nullable but shouldn't. 

Fixes: https://github.com/supabase/supabase-js/issues/1368


test/select-query-parser/select.test-d.ts
```diff
@@ -85,6 +85,20 @@ type Schema = Database['public']
...
+// !inner relationship on nullable relation
+{
+  const { data } = await selectQueries.innerJoinOnNullableRelationship
+  let result: Exclude<typeof data, null>
+  let expected: Array<{
+    id: number
+    hotel: {
+      id: number
+      name: string | null
+    }
```
avallete, line 97: **note**

Since this come from an `!inner` join it should not be null as it's filtered at the join level.

soedirgo (approved):

PR #613, opened 2025-03-29
chore(tests): add auto types gen and override for testing

avallete:

This PR improve the testing flow to make it more e2e by:

1. Automatically regenerating the types definition for the test database from the latest `postgres-meta` version
2. Using the "type override" for Json fields that we want to test the override on. Like recommended in the documentation. Giving an environment closer to what an user project will look like.

This will allow us to more easily go from a "sql definition" to an "type inference test".


package.json
```diff
@@ -39,10 +39,12 @@
...
     "test": "run-s format:check test:types db:clean db:run test:run db:clean && node test/smoke.cjs && node test/smoke.mjs",
     "test:run": "jest --runInBand --coverage",
-    "test:update": "run-s db:clean db:run && jest --runInBand --updateSnapshot && run-s db:clean",
+    "test:update": "run-s db:clean db:run db:generate-test-types && jest --runInBand --updateSnapshot && run-s db:clean",
     "test:types": "run-s build && tsd --files 'test/**/*.test-d.ts'",
+    "test:types:watch": "run-s build && tsd --files 'test/**/*.test-d.ts' --watch",
     "db:clean": "cd test/db && docker compose down --volumes",
-    "db:run": "cd test/db && docker compose up --detach && wait-for-localhost 3000"
+    "db:run": "cd test/db && docker compose up --detach && wait-for-localhost 3000",
+    "db:generate-test-types": "cd test/db && docker compose up --detach && wait-for-localhost 8080 && curl --location 'http://0.0.0.0:8080/generators/typescript?included_schemas=public,personal&detect_one_to_one_relationships=true' > ../types.generated.ts && sed -i 's/export type Json = .*/export type Json = unknown;/' ../types.generated.ts"
```
avallete, line 47: **note**

Here we replace the `Json` type definition by `unkown` as it allow `MergeDeep` to work better to override the column type with a custom definition.

Long term, we will deprecate our custom type in favor of unknown for all Json types including for the `select-query-parser`. This might be a breaking change in the types system for some users.

Related: https://github.com/supabase/postgres-meta/pull/750

test/select-query-parser/result.test-d.ts
```diff
@@ -1,9 +1,11 @@
-import { Database, Json } from '../types'
+import { Database } from '../types.override'
 import { selectParams } from '../relationships'
 import { GetResult } from '../../src/select-query-parser/result'
 import { expectType } from 'tsd'
 import { TypeEqual } from 'ts-expect'
 import { SelectQueryError } from '../../src/select-query-parser/utils'
+// TODO: should change this type in favor of unknown instead of the current one
+import type { Json } from '../../src/select-query-parser/types'
```
avallete, line 8: **note**

Since `select-query-parser` has it's own static Json type for json field selectors, we must use this one for type equality comparator.

soedirgo (approved):

PR #614, opened 2025-03-30
feat(types): add inference for embeded joins by functions

avallete:

Once https://github.com/supabase/postgres-meta/pull/915 is merged

This introduce automatic types inferences for "joins based on functions" to `postgrest-js`.

Along with the type generation in `pg-meta` it should allows queries such as:

```ts
postgrest
        .from('organizations')
        .select('subscription_id, custom_function_returning_setof_other_table(customer_id)')
```

To work out of the box.

It also introduce a way to "force cast" an object relation as being not nullable at the database override level in case some DB layer app logic enforce it.

This should not be a breaking change since everything has be scoped in an additional type.

Fixes: CLIBS-79

Related: https://github.com/supabase/supabase-js/issues/1259

Edit: Tested the full pipeline (typegen + inference) over our infra repo, was able to get rid of all types override for the `billing_subscriptions` (see: https://github.com/supabase/infrastructure/compare/develop...chore/upgrade-supabase-js)


src/PostgrestTransformBuilder.ts
```diff
@@ -23,7 +24,7 @@ export default class PostgrestTransformBuilder<
     NewResultOne = GetResult<Schema, Row, RelationName, Relationships, Query>
   >(
     columns?: Query
-  ): PostgrestTransformBuilder<Schema, Row, NewResultOne[], RelationName, Relationships> {
+  ): PostgrestFilterBuilder<Schema, Row, NewResultOne[], RelationName, Relationships> {
```
avallete, line 27: **note**

This should fix: https://github.com/supabase/supabase-js/issues/1365

test/types.override.ts
```diff
@@ -43,6 +43,13 @@ export type Database = MergeDeep<
           }
         }
       }
+      Functions: {
+        get_user_profile_non_nullable: {
+          SetofOptions: {
+            isNotNullable: true
```
avallete, line 49: **note**

This will be the api to make any row returning function globally "non nullable" if desired by the user, to avoid having to use the `!` operator everytime.

gwax:
Is there any progress on this?

avallete:
Closes in favor of: https://github.com/supabase/postgrest-js/pull/632

PR #618, opened 2025-05-19
feat(types): setup postgrest-13 types inference

avallete:


src/select-query-parser/result.ts
```diff
@@ -392,21 +428,46 @@ type ProcessEmbeddedResourceResult<
...
+  Spread['target']
+> extends infer Result
   ? Result extends SelectQueryError<infer E>
     ? SelectQueryError<E>
     : ExtractFirstProperty<Result> extends unknown[]
-    ? {
-        [K in Spread['target']['name']]: SelectQueryError<`"${RelationName}" and "${Spread['target']['name']}" do not form a many-to-one or one-to-one relationship spread not possible`>
-      }
+    ? ClientOptions['postgrestVersion'] extends 13 // Spread over an many-to-many relationship, turn all the result fields into arrays
+      ? ProcessManyToManySpreadNodeResult<Result>
```
avallete, line 449: **note**

Here is what the conditionality for the inference can look like.

test/db/docker-compose.yml
```diff
@@ -2,7 +2,21 @@
...
+      PGRST_DB_SCHEMAS: public,personal
+      PGRST_DB_EXTRA_SEARCH_PATH: extensions
+      PGRST_DB_ANON_ROLE: postgres
+      PGRST_DB_PLAN_ENABLED: 1
+      PGRST_DB_TX_END: commit-allow-override
+      PGRST_DB_AGGREGATES_ENABLED: true
+    depends_on:
+      - db
+  rest12:
     image: postgrest/postgrest:v12.2.0
```
avallete, line 20: **note**

It is also possible to check if all the current tests and the new version ones are matching by upgrading the `rest12` container to `postgrest:v13`. I did so for testing, and except from the `Spread on Many` feature, everything was behaving the same, with the only difference being the returned error message within some of the inline snapshots.

src/select-query-parser/result.ts
```diff
@@ -392,21 +428,46 @@ type ProcessEmbeddedResourceResult<
...
+  Relationships,
+  Spread['target']
+> extends infer Result
   ? Result extends SelectQueryError<infer E>
     ? SelectQueryError<E>
     : ExtractFirstProperty<Result> extends unknown[]
-    ? {
-        [K in Spread['target']['name']]: SelectQueryError<`"${RelationName}" and "${Spread['target']['name']}" do not form a many-to-one or one-to-one relationship spread not possible`>
-      }
+    ? ClientOptions['postgrestVersion'] extends 13 // Spread over an many-to-many relationship, turn all the result fields into arrays
```
steve-chavez, line 448: Q: Will this comment appear on the docs? If so, it would be good to mention the arrays are correlated (they share the same order).

```suggestion
    ? ClientOptions['postgrestVersion'] extends 13 // Spread over an many-to-many relationship, turn all the result fields into correlated arrays
```

avallete: Not really this comment is only there for devs to understand why we do branch here and the behavior this type cover. Sounds good to mention the correlation regardless 👍

src/PostgrestBuilder.ts
```diff
@@ -8,12 +8,16 @@ import type {
...
   IsValidResultOverride,
+  ClientServerOptions,
 } from './types'
 import PostgrestError from './PostgrestError'
 import { ContainsNull } from './select-query-parser/types'
 
-export default abstract class PostgrestBuilder<Result, ThrowOnError extends boolean = false>
-  implements
+export default abstract class PostgrestBuilder<
+  ClientOptions extends ClientServerOptions,
```
soedirgo, line 17: Won't block the PR based on this but I have a hunch that adding new type parameters at the start/middle of the parameter list will break some 3rd party lib relying on the typings

avallete: Hum that's true, I've looked for usage on github: https://github.com/search?q=language%3ATypeScript++PostgrestBuilder%3C&type=code&p=3

Most of them seems to be forks though. Maybe we could pass this into a new major version though.

Or maybe in the changelog we can give a fix solution for compatibility:

```ts
import type { PostgrestBuilder as BasePostgrestBuilder } from '@supabase/posgrest-js';

type PostgrestBuilder<R, TE extends boolean = false> = BasePostgrestBuilder<{}, R, TE>
```

Didn't tested it but this should allow to use the `PostgrestBuilder` type as before 🤔 

I think I remember having trouble pushing this as a last argument because then, it required to always also define the `ThrowOnError` param which has default set.

soedirgo (approved):

PR #619, opened 2025-05-19
feat: postgrest 13 add maxaffected in client libraries

avallete:

- Implement the [maxAffected](https://docs.postgrest.org/en/v13/references/api/preferences.html#max-affected) postgrest 13 feature at the client level
- Add tests, introduce a more generic way to handle Prefer header, can be refactor at other places in other PR


test/max-affected.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    // Try to update all rows with maxAffected=2
+    const result = await postgrest13
+      .from('messages')
+      .update({ message: 'updated' })
+      .eq('message', 'test1')
+      .maxAffected(2)
+    const { error } = result
+    expect(error).toBeDefined()
+    expect(error?.message).toBe('Query result exceeds max-affected preference constraint')
```
steve-chavez, line 78: How about just matching the PostgREST error code? (`PGRST124`, [ref](https://docs.postgrest.org/en/v13/references/api/preferences.html#max-affected))

This would allow us to change the error message without causing a test failure later. Right now the error message seems a bit specific to postgREST ("preference constraint" could be removed)

test/max-affected.ts
```diff
@@ -0,0 +1,138 @@
...
+    // Try to delete all rows with maxAffected=2
+    const { error } = await postgrest13
+      .from('messages')
+      .delete()
+      .eq('message', 'test3')
+      .maxAffected(2)
+      .select()
+
+    expect(error).toBeDefined()
+    expect(error?.message).toBe('Query result exceeds max-affected preference constraint')
```
steve-chavez, line 117: ditto here, only match error code if possible.

src/utils.ts
```diff
@@ -0,0 +1,64 @@
+export class HeaderManager {
```
soedirgo, line 1: Can we use the [Headers API](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instead? Should be available on both web and Node

soedirgo: Great idea though and we should move to using that vs. plain objects/`Record`s

src/PostgrestFilterBuilder.ts
```diff
@@ -599,4 +602,28 @@ export default class PostgrestFilterBuilder<
...
     return this
   }
+
+  /**
+   * Set the maximum number of rows that can be affected by the query.
+   * Only available in PostgREST v13+ and only works with PATCH and DELETE methods.
+   *
+   * @param value - The maximum number of rows that can be affected
+   */
+  maxAffected(
```
soedirgo, line 612: Let's move this to `PostgrestTransformBuilder` since this doesn't do horizontal filtering, but rather an assertion a la `.single()`

src/utils.ts
avallete: That's quite the refactor, it's scoped there: https://github.com/supabase/postgrest-js/pull/619/commits/fe3e80f8cd69224da36dd2094284cbf9c1e7dc98

avallete: This also require to bump minimal node version to 18 see: [549abdf](https://github.com/supabase/postgrest-js/pull/619/commits/549abdfc0758f6d52cf90652c7b69ce83b8755b3)

LMK if we're comfortable doing that.

soedirgo: Ah 😂 I meant we could refactor that in a future PR. Thanks for getting that done!

src/PostgrestTransformBuilder.ts
```diff
@@ -344,4 +339,26 @@ export default class PostgrestTransformBuilder<
...
+  /**
+   * Set the maximum number of rows that can be affected by the query.
+   * Only available in PostgREST v13+ and only works with PATCH and DELETE methods.
+   *
+   * @param value - The maximum number of rows that can be affected
+   */
+  maxAffected(
+    value: number
+  ): ClientOptions['postgrestVersion'] extends 13
+    ? Method extends 'PATCH' | 'DELETE'
```
steve-chavez, line 352: @avallete This works for RPC too, could we add it that? The docs have been updated recently ([ref](https://docs.postgrest.org/en/v13/references/api/preferences.html#max-affected), see RPC at the bottom).

avallete: Good catch didn't tested the `rpc` they do have their own kind of typing though because they're a bit specifics.

About the "set of" that's actually something I wanna work on. It will be related to "embedded relationships with functions". I have some draft I need to find the time to work on it.

steve-chavez: > For now I think runtime error is okay, but we'll surely be able to catch that at type level at some point indeed. Just wondering, what if the function doesn't returns a setof but just plain records ? Or return table ?

Right now it will just fail silently, but we're working on a fix to raise an error.

avallete: Added a specific case for `RPC` calls as we can't really know the request method at type inference time, nor can we know if it returns a \`setof\` another table. So it'll be a bit of degraded experience typewise (more chances to have error only at runtime) but should be good enough for now.

steve-chavez: > Right now it will just fail silently, but we're working on a fix to raise an error.

JFYI, the above is now fixed and now an error will be raised: https://github.com/PostgREST/postgrest/releases/tag/v13.0.3

soedirgo (approved):

PR #621, opened 2025-07-17
chore: refactor tests

avallete:

- Update tests snapshots
- Include types tests within the runtime tests and inline more for better readability between what's tested and the expected results.
- Rename all runtime tests `*.test.ts`
- Refactor some of the tests types to deduplicate and split them up in files by scope of what's tested.
- Split down large tests into smaller units
- Use testSequencer to ensure tests are always run in the same order (since the "write" tests side-effect the other results otherwise).


avallete:
> Can we split these into separate PRs? This is a bit difficult to review

Yup, can do, started by merging both the shared types/runtime tests together into the same one to reduce context switching here: https://github.com/supabase/postgrest-js/pull/622

Will sequence the rest to minimize diffs for easier reviews.

PR #623, opened 2025-07-18
chore: rename and add testSequencer

avallete:

- Rename all tests with `.test.ts` extension
- Add a testSequencer to ensure deterministic order since `index.test.ts` is not there anymore to ensure it. (TODO: would be great for the tests to not have side effects between each-others)
- Update snapshots


test/relationships.test.ts
```diff
@@ -38,6 +38,42 @@ test('nested query with selective fields', async () => {
...
+            "message": "test3",
+          },
+          Object {
+            "id": 17,
+            "message": "test3",
+          },
+          Object {
+            "id": 18,
+            "message": "test3",
+          },
```
soedirgo, line 76: Let's delete these rows where they were inserted? One test should affect the others

soedirgo (approved):

PR #625, opened 2025-07-18
chore: split basics test

avallete:

- Split basic.test.ts into smaller units, review and deduplicates duplicates
- Update snapshots since file order changed with new added files


avallete:
Will re-work this one when got more bandwith for it.

PR #626, opened 2025-07-21
fix(types): computed field and star selector

avallete:

- When using star selector, postgrest will exclude computed field from the automatic selection, and only add those if explicitly requested in the select query


test/select-query-parser/types.test-d.ts
```diff
@@ -53,3 +53,282 @@ import { DeduplicateRelationships } from '../../src/select-query-parser/utils'
   type result = DeduplicateRelationships<rels>
   expectType<TypeEqual<result, expected>>(true)
 }
+// Test GetComputedFields basic
+{
+  type Json = unknown
+
+  type Database = {
```
soedirgo, line 60: Can we reuse the generated types instead of using a mock?

soedirgo (approved):

PR #627, opened 2025-07-28
chore(tests): setup zod type checking unify

avallete:

Still in the idea to make a more straightforward, unified and strongly typed values for our tests: 

- Introduce `zod` schemas and re-use the same schemas to both:
  - infer the expected result type
  - Ensure that the runtime result type match the validation of the schema

- Doing so, found a bug where self-referencing embeding via columns name would cause an unwanted intersection and fixed it.

### DIFF
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -1,0 +1,12 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["src", "test"],
+  "compilerOptions": {
+    "strict": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+    "skipLibCheck": true,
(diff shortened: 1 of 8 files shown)


tsconfig.test.json
```diff
@@ -0,0 +1,12 @@
+{
```
soedirgo, line 1: Any reason why this is needed? Seems the tests work without it

avallete: Mainly because without this I would have difference between what's VSCode considered like an error an the result I would get out of `type-check`. With this being explicitly set, I get the same consistent result between the VSCode editor and the type-check results when running the tests with `jest`.

PR #630, opened 2025-08-04
fix(typegen): avoid possible infinite recursion error

avallete:
- Fix #627 regression with `Omit`
- Add test to ensure that the provided fix remove this error from a recursive embeding

Related:
https://github.com/supabase/supabase-js/issues/1354
https://github.com/supabase/supabase-js/issues/1372
https://github.com/supabase/supabase-js/issues/808


package.json
```diff
@@ -40,8 +40,8 @@
     "test": "run-s format:check test:types db:clean db:run test:run db:clean && node test/smoke.cjs && node test/smoke.mjs",
     "test:run": "jest --runInBand --coverage",
     "test:update": "run-s db:clean db:run db:generate-test-types && jest --runInBand --updateSnapshot && run-s db:clean",
-    "test:types": "run-s build && tsd --files 'test/**/*.test-d.ts'",
-    "test:types:watch": "run-s build && tsd --files 'test/**/*.test-d.ts' --watch",
+    "test:types": "run-s build && tsd --files 'test/**/*.test*.ts'",
```
avallete, line 43: Now that we have types and tests merged together, we want to check for all tests files that the types assertions passes.

package.json
```diff
@@ -40,8 +40,8 @@
...
+    "test:types:watch": "chokidar 'src/select-query-parser/**/*.ts' 'test/**/*.ts' -c 'npm run test:types'",
```
avallete, line 44: Previous watcher weren't working. Use chodikar for better DX to check the types errors.

avallete (commented):
After attempting many things:

(based on: https://www.esveo.com/en/blog/how-to-workaround-the-max-recursion-depth-in-typescript/)
- Tail recursion
- Batches handling
- Reseting counter to 0

None of them really improved the issue,  I've figured out that this error is mainly a regression due to the newly introduced `Omit`. It'll still happen for complex queries though.

The only "optimization" that had a bit of value were hardcoding some "batches" to reduce the recursion inference. It didn't allowed much deeper recursions though, only improved a bit the inference speed (~10%). Might introduce this in future changes, keeping this one focused on reversing the regression and adding tests cases for it.

PR #633, opened 2025-08-07
test(types): introduce tstyche instead of tsd

avallete:

Following up: https://github.com/supabase/postgrest-js/pull/632#discussion_r2260796038

- Introduce `tstyche` to improve the type testing by adding expected message matching to the `@ts-expect-errors`
- Leverage `tstyche` to check that types tests passes across a wide ranges of Typescript versions (`>=4.7`)
- Remove `tsd` since we no longer needs it. Remove `ts-expect` and extract `TypeEqual` which is the only used utils from it.

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -1,0 +1,167 @@
+name: Preview release
+
+permissions:
+  pull-requests: write
+
+on:
+  # Manual trigger with inputs for control
+  workflow_dispatch:
+    inputs:
+      trigger_supabase_js:
+        description: 'Trigger supabase-js tests'
+        type: boolean
+        default: true
+      target_branch:
+        description: 'Target branch for supabase-js tests'
+        type: string
+        default: 'master'
+  
+  # Push to master - only when source code changes
+  push:
+    branches:
+      - master
+    paths:
+      - 'src/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'tsconfig.json'
+  
+  # PR triggers - only when labeled
+  # Using pull_request_target to access secrets when PRs come from forks
+  pull_request_target:
+    types: [labeled, synchronize]
+
+jobs:
+  preview:
+    # Run only for PRs with 'trigger: preview' label or pushes to master
+    if: >
+      github.repository == 'supabase/postgrest-js' &&
+      (
+        github.event_name == 'workflow_dispatch' ||
+        github.event_name == 'push' ||
+        (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'trigger: preview'))
+      )
+    runs-on: ubuntu-latest
+    outputs:
+      preview-url: ${{ steps.preview.outputs.url }}
+      package-name: ${{ steps.preview.outputs.package }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          # For pull_request_target, we need to explicitly checkout the PR's head
+          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Publish preview
+        id: preview
+        run: |
+          OUTPUT=$(npx pkg-pr-new@latest publish --compact 2>&1)
+          PREVIEW_URL=$(echo "$OUTPUT" | grep -o 'https://pkg\.pr\.new/@supabase/[^[:space:]]*' | head -1)
+          REPO_NAME=$(echo "$GITHUB_REPOSITORY" | cut -d'/' -f2)
+
+          if [ -z "$PREVIEW_URL" ]; then
+            echo "Error: Failed to extract preview URL from pkg-pr-new output"
+            echo "Output was: $OUTPUT"
+            exit 1
+          fi
+
+          echo "Preview Release URL: $PREVIEW_URL"
+          echo "url=$PREVIEW_URL" >> $GITHUB_OUTPUT
+          echo "package=$REPO_NAME" >> $GITHUB_OUTPUT
+
+  trigger-supabase-js-tests:
+    needs: preview
+    # Only run if preview URL exists and either:
+    # - Not workflow_dispatch, OR
+    # - workflow_dispatch with trigger_supabase_js = true
+    if: >
+      needs.preview.outputs.preview-url != '' &&
+      (
+        github.event_name != 'workflow_dispatch' ||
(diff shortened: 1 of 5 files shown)


test/types.ts
```diff
@@ -0,0 +1,5 @@
+import { expect } from 'tstyche'
+
+export function expectType<T>(expression: T) {
+  return expect(expression).type.toBe<T>()
+}
```
avallete, line 5: **comment**

Just use this to keep the nice `tsd` DX where types mismatch would shows up in code editor but still be able to remove the `tsd` dependecy itself and migrate to `tstyche` instad.

Should no longer be necessary if: https://github.com/tstyche/tstyche/issues/467 lands.

test/types.ts
```diff
@@ -0,0 +1,5 @@
+import { expect } from 'tstyche'
+
+export function expectType<T>(expression: T) {
+  return expect(expression).type.toBe<T>()
```
mrazauskas, line 4: Are you sure this works? Types only statically analysed, so this would pass because type of `expression` is `T`. All other checks would only check assignability in `expectType<T>(expression: T)` level without reaching this line.

test/types.ts
mrazauskas: I do understand the feeling of seeing the red squiggles in the editor. But.. the `tsd`s `expectType()` gives in-editor feedback only for assignability. The identically relation is checked programatically: https://github.com/tsdjs/tsd/blob/main/source/lib/assertions/handlers/identicality.ts

At the same time, the `tsd`s `expectNotType()` does not show the red squiggles ever. This mismatch in implementation makes assertions with and without `Not` to pass with same values. https://github.com/tsdjs/tsd/issues/141

These are two reasons why there are no red squiggles in TSTyche. The old good separation of concerns. I do plan to improve DX by having editor extension / plugins some time later. 

---

As a test example, try:

```ts
expectType<{ a: string | number }>({ a: "test" })
```

It gives no squiggles, but `tsd` would fail the test. Same with `expect().type.toBe()`, but without any squiggles and bugs (;

test/issue-1354-d.ts
mrazauskas, line 1: This file is not picked up by the `testFileMatch` patterns. Could be renamed, perhaps?

test/types.ts
mrazauskas: On second thought, it works fine when `expectType()` is used in combination with `TypeEqual` from `ts-expect`. The following shows error in the editor:

```ts
expectType<TypeEqual<{ a: string | number }, { a: string }>>(true)
```

---

The notable difference is output in CLI (for example, in CI environment).

With `expectType()`:

Image

With TSTyche assertions:

Image

mrazauskas (commented):
Thanks for giving TSTyche a try. Glad to see this. I have left few inline comments.

One more important detail. By default, TSTyche logs the installed `typescript` package. The version used for testing is printed in the log:

Image

This is different from `tsd` which ships with its own copy of `@tsd/typescript`. The exact version can be found here: https://github.com/supabase/postgrest-js/blob/823f28f2a3927c2dead02e3e9ce8429c6de60afb/package-lock.json#L1123-L1124

---

All depends on your intention. You can run `npx tstyche --target 5.4` to test against TypeScript 5.4.5. (Few checks error messages are failing because of minor changes between TypeScript versions.)

Or you can run `npx tstyche --target '>=5.4'` to test against TypeScript 5.4, 5.6, 5.7, etc. (Takes time. Might be too much for local checks. But definitely good idea for CI.)

mrazauskas:
> So `tstyche --target '>=4.5'`.

Only `>=4.7` is supported. Sorry... But if you run `npx tsc --version` here or in `supabase-js` repo, they both are currently using TypeScript 4.7.4. So, perhaps `tstyche --target '>=4.7'` is fine?

---

- TypeScript 4.7 was released 3 years ago ([source](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/))
- DefinitelyTyped tests types agains versions of TypeScript that are less than 2 years old ([source](https://github.com/DefinitelyTyped/DefinitelyTyped?tab=readme-ov-file#support-window))

That’s why I was thinking that keeping support only for `>=4.7` is fine. Also TypeScript 4.7 was the first one with modernised module resolution (`node16`). This means compiler options got modernised and testing older versions got more complicated, because the new options are causing errors.

avallete:
> > So `tstyche --target '>=4.5'`.
> 
> Only `>=4.7` is supported. Sorry... But if you run `npx tsc --version` here or in `supabase-js` repo, they both are currently using TypeScript 4.7.4. So, perhaps `tstyche --target '>=4.7'` is fine?
> 
> * TypeScript 4.7 was released 3 years ago ([source](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/))
> * DefinitelyTyped tests types agains versions of TypeScript that are less than 2 years old ([source](https://github.com/DefinitelyTyped/DefinitelyTyped?tab=readme-ov-file#support-window))
> 
> That’s why I was thinking that keeping support only for `>=4.7` is fine. Also TypeScript 4.7 was the first one with modernised module resolution (`node16`). This means compiler options got modernised and testing older versions got more complicated, because the new options are causing errors.

Still worth mentioning that re-writing the types, we did had some frictions using some non `4.5` compatibles typing logic: https://github.com/supabase/postgrest-js/pull/571

But as time pass, I guess we can expect people to migrate more and more and this retro-compatiblity to be less and less useful.

package.json
```diff
@@ -38,10 +38,12 @@
     "docs": "typedoc src/index.ts --out docs/v2",
     "docs:json": "typedoc --json docs/v2/spec.json --excludeExternals src/index.ts",
     "test": "run-s format:check test:types db:clean db:run test:generated-types test:run db:clean && node test/smoke.cjs && node test/smoke.mjs",
+    "test:ci": "run-s format:check test:types:ci db:clean db:run test:generated-types test:run db:clean && node test/smoke.cjs && node test/smoke.mjs",
     "test:run": "jest --runInBand --coverage",
     "test:generated-types": "run-s db:generate-test-types && git diff --exit-code test/types.generated.ts || (echo '❌ Generated types are not up to date. Run npm run db:generate-test-types.' && exit 1)",
     "test:update": "run-s db:clean db:run db:generate-test-types && jest --runInBand --updateSnapshot && run-s db:clean",
-    "test:types": "run-s build && tsd --files 'test/**/*.test*.ts'",
+    "test:types": "run-s build && tstyche",
+    "test:types:ci": "run-s build && tstyche --target '>=4.7'",
```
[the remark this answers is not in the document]
avallete: Changed toward specifics versions, testing them all were doubling CI time while not providing much more coverage.

Now we check over `4.7,5.5,latest`.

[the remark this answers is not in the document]
avallete: Maybe parallelism between versions, but I wonder how hard it is to implement, I guess each Typescript version could cause conflict with the next one 🤔

soedirgo (approved):

test/select-query-parser/parser.test-d.ts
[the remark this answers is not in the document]
soedirgo: Good catch!

PR #648, opened 2025-09-30
Feat/expose query builder options

stenlan:

Introduces a way to pass options to the PostgrestQueryBuilder instance on a per-request basis.

There is no way to specify per-request options currently.
https://github.com/supabase/supabase-js/issues/438

## What is the new behavior?

An optional parameter is introduced that can be passed to the `.from` method, where users can optionally specify custom `fetch` and `headers` options.


mandarini:
Thank you very much for your time and contribution @stenlan ! I would really appreciate if you could move these changes in the new repo: https://github.com/supabase/supabase-js

Closing this PR now, as per comment here: https://github.com/supabase/supabase-js/pull/1570#issuecomment-3385802483

